PR #1, opened 2020-11-30
Import of keg recipes

jgleissner:
Import of everything I've produced so far. I haven't bothered with OCI profiles at this point, but otherwise the recipes should cover all SLES 15 SP2 images. I haven't verified yet if the generated image descriptions completely match the existing ones in content.

Obviously this is work in progress but should hopefully be useful for discussion. I've included a README that explains the basics, this should be replaced by proper documentation in the future. I've also included a simple python script `keg-tester`, which can be used to produce kiwi files from the recipes. This should be removed once we have the actual keg tool, of course, but for now it's useful for testing. Example usage (in project root): ./keg-tester sles/ondemand/15-sp2


data/base/jeos/sle15/packages.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - audit
+      - bc
+      - blog
+      - cifs-utils
+      - dosfstools
+      - dracut
+      - glibc-locale
+      - grub2
+      - name: grub2-x86_64-efi
+        arch: x86_64
```
schaefi, line 14: Is this by convention ? I wondered why e.g grub2 is just a list member when grub2-x86_64-efi is a list member that is a hash. I guess this is done to allow other attributes like arch: right ?

data/base/sles/sle15/config.yaml
```diff
@@ -0,0 +1,34 @@
+config:
+  SLES:
+    sysconfig:
```
schaefi, line 3: This information will be used to create config.sh correct ? As config.sh is a custom script that can contain anything do we have an idea how raw scripting should be represented in yaml. I'm asking this because raw data in yaml and also json has some pitfalls e.g on line breaks

So how would we represent this script part as example:

```
pushd /etc/products.d
ln -sf SLES_SAP.prod baseproduct
popd
```

data/csp/ec2/ondemand/sle15/packages.yaml
```diff
@@ -0,0 +1,6 @@
+packages:
+  image:
+    EC2 ondemand:
```
schaefi, line 3: do we want to allow spaces in hash keys ?

keg-tester
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/python3
+
```
schaefi, line 2: I assume this part will go into the keg repo and into a "real" implementation ?

schaefi (approved):
Overall looks good :+1: quite some effort, wow

This is all work in progress so I looked at it and responded whenever I thought this might be an issue later or when I had a question

One part I think should be changed is that the complete set of information including the templates should be stored under a directory/namespace that indicates that this targets public cloud providers. e.g

```
pubcloud/schemas/...
pubcloud/images/...
pubcloud/data/...
```

Thanks

data/csp/ec2/ondemand/sle15/packages.yaml
rjschwei: I think not. If the snippet lives in a tree that is already labeled with `ec2` do we need the code to call this out again as in `ec2-on-demand` or would `on-demand` be sufficient? For example for the profile things are not labeled as `azure-profile` or `ec2-profile`.

data/products/sap/sle15/packages.yaml
```diff
@@ -0,0 +1,47 @@
+packages:
+  bootstrap:
+    SAP:
+      - sle-ha-release
+      - sle-module-live-patching-release
+      - sle-module-sap-applications-release
```
rjschwei, line 6: duplicates `data/products/sap/ondemand/sle15/packages.yaml`, I am not certain I understand the slicing between BYOS and on-demand in this respect.

rjschwei (changes requested):
It's a great start, also a lot to digest, I am not certain I caught everything and we have a couple of items that need additional clarification.

jgleissner:
> One part I think should be changed is that the complete set of information including the templates should be stored under a directory/namespace that indicates that this targets public cloud providers. e.g
> 
> ```
> pubcloud/schemas/...
> pubcloud/images/...
> pubcloud/data/...
> ```
I think if we only intend to host public cloud specific recipes in this repository then we wouldn't really need it. And some of the data could actually be used for other image types.

data/base/jeos/sle15/packages.yaml
jgleissner: Yes, I needed a way to optionally specify the arch flag, but I didn't like having the name tag everywhere, so I added the option to specify the package name as a simple string, and the schema has a check for that. It maybe looks a bit odd.

data/base/sles/sle15/config.yaml
jgleissner: The `config` section isn't used yet, this is basically a left-over from the initial example. I'll work on it next.

data/csp/ec2/ondemand/sle15/packages.yaml
jgleissner: I don't think spaces should be a problem in keys, but I'm happy to get rid of them. I would keep the EC2 reference though, as it reduces the risk of namespace collision. Also, this way you see where the package selection comes from (it even gets added as a comment in the generated config.kiwi).

keg-tester
jgleissner: Yes, see initial comment.

data/base/registration/sle15/packages.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - sle-module-basesystem-release
+      - sle-module-cap-tools-release
+      - sle-module-containers-release
+      - sle-module-desktop-applications-release
+      - sle-module-development-tools-release
+      - sle-module-legacy-release
+      - sle-module-public-cloud-release
+      - sle-module-python2-release
+      - sle-module-server-applications-release
+      - sle-module-web-scripting-release
```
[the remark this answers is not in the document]
jgleissner: Naming is probably not idea, agreed. The packages in registration are only in on-demand images not in byos images, but they're not csp specific that's why included them in the base branch rather than csp.

data/products/sap/sle15/packages.yaml
jgleissner: Yes, this should only be in ondemand, my bad. Wrt slicing, I only replicated what's in the current image descriptions. The release packages of the various modules are included in the on-demand version but not in the BYOS ones. It's probably a good idea to not include release packages in BYOS images, as they should only be installed when the according module is enabled. Not sure whether including them in on-demand images is necessary. I would expect them to be installed automatically when the modules get enabled during registration.

README.md
```diff
@@ -0,0 +1,92 @@
...
+may scan additional sub directories based on the `include-suffix` parameter of
+the image recipe, and this applies to all parent directories as well. This can
+be used to add OS version specific changes to the generic configuration.
+
+To visualize how this works in practice, let's look at the configuration sets
+for EC2 support. A recipe for a SLES 15 SP2 EC2 image may contain the following
+configuration parameters:
+
+```
+include-prefix: sle15/sp2
```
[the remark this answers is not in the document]
jgleissner: The current layout only inherits from the layers above, so  sle15/sp2 would inherit from sle15 only. The idea is that we have the common definitions in sle15 and only add sp specific snippets where necessary, which should hopefully be rare.

Said that, multiple include-prefix (which really should be include-suffix, or maybe a better name altogether) could be an option. Not currently supported by keg-tester, but should be feasible.

rjschwei:
> > One part I think should be changed is that the complete set of information including the templates should be stored under a directory/namespace that indicates that this targets public cloud providers. e.g
> > ```
> > pubcloud/schemas/...
> > pubcloud/images/...
> > pubcloud/data/...
> > ```
> 
> I think if we only intend to host public cloud specific recipes in this repository then we wouldn't really need it. And some of the data could actually be used for other image types.

I agree. I think the slicing along target lines as we already have it should be sufficient. We do want 3rd parties to use our setup and if they build something for their on-premise cloud they should be able to build from the target, i.e. "VM build" and should not be forced to pull from something named "pubcloud". In short I think the template slicing per build target is sufficient.

data/base/jeos/sle15/packages.yaml
rjschwei: I think it's obvious enough and nicer than inventing syntax such as `- grub2-x86_64-efi:x86_64`. This would make it look more consistent in the snippet but opens the door to other problems. The "odd" look does not bother me.

README.md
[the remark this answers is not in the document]
rjschwei: The concrete example here is the grub timer setting that was introduced in grub for 15 SP1 and supported in kiwi. If we cannot inherit across SPs then we need to repeat the snippet that sets the grub timer behavior for 15 SP1 & 2 & 3 & 4 &5 & 6, but if we can inherit from any lower version then we only need to add the data to 15 SP1 and it is used for everything coming later. That's the approach used in mashhouse and I think will serve us well here.

data/base/registration/sle15/packages.yaml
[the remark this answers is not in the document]
rjschwei: No problem with the location in the base, this is a slicing topic in that the modules get added to on-demand. So yes the naming/slicing needs to be fixed.

data/products/sap/sle15/packages.yaml
rjschwei: The registration is driven of the included product packages which is why the release packages are in the image builds for on-demand. The classification as provided by SCC/RMT as to product <-> module relationship and what may and may not be enabled by default is exceedingly complicated and may lead to a "not mirrored" issue.

schaefi:
>I think if we only intend to host public cloud specific recipes in this repository then we wouldn't really need it. And some of the data could actually be used for other image types.

I agree if that repo is only intended to hold our stuff. To be consistent I suggest to rename the repo into ```keg-pubcloud-recipes``` because the current name implies generic use which I also assumed when I wrote the comment. Thoughts ?

rjschwei:
> > I think if we only intend to host public cloud specific recipes in this repository then we wouldn't really need it. And some of the data could actually be used for other image types.
> 
> I agree if that repo is only intended to hold our stuff. To be consistent I suggest to rename the repo into `keg-pubcloud-recipes` because the current name implies generic use which I also assumed when I wrote the comment. Thoughts ?

Sorry I do not follow you. We will maintain our stuff here. If 3rd party users want to use keg-recepies and contribute here they are welcome to do so. Meaning if one want to add a `netboot` tree and maintain those things here that is perfectly reasonable. We already agreed that we'd have a slice across build targets, so I do not see where `pubcloud` is different from `vm`. Where the `vm` designator catches the various formats `raw`, `qcow2` `vhdifxed` etc, But I think I am missing the point you are trying to make.

schaefi:
I've looked over the data again and think you are right. My point was that in the amount of data I did not saw that we put pubcloud specific stuff under a namespace. But we do as for example in ```data/csp/azure/sle15/config.yaml```. Ok so forget what I said. I think this is good. It's just hard to review :) Sorry for the noise

README.md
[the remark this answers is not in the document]
jgleissner: I've pushed some changes that allow to specify multiple include paths so you can have an image definition that inherits from other paths (i.e. service packs). I've also added a couple of SLES 15 SP1 and GA definitions to showcase that, and also to cover the point in my last comment (I added an option to remove a previously set parameter by setting it to "None"; come to think of it, maybe something like "_clear" or "_unset" would be better).

jgleissner:
I've added a singebuild template specifically for singlebuild images, which allows to have the single- and multibuild definitions living side-by-side. It's perhaps not idea because there is duplication in the templates. I intend to move some common functionality into a separate file, but to allow templates to load other templates you need to define a loading mechanism and I didn't bother to implement that in keg-tester. Once we go ahead with the structure and start implementing the tool properly this can be tackled, but for now there's that duplication in the schemas.

data/base/registration/sle15/packages.yaml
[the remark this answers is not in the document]
jgleissner: I renamed it to `cloudreg` to make it somewhat clearer that this is for cloudregister functionality not for wider SLES registration.

data/csp/ec2/ondemand/sle15/packages.yaml
rjschwei: Well it's about consistency and duplication. I am not certain how `ec2-on-demand` is different from `ec2-profile`, `azure-profile` etc. But for the profile it was chosen not to have such a prefix. I don't think we should block this initial PR on the naming conversation but think we should look at this again.

data/base/registration/sle15/packages.yaml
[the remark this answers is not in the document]
rjschwei: We still have the -release packages in this snippet. But release packages get installed in `bootsrap` to avoid IBS related unpleasantness with -mini packages and the actual registration code is part of `images`. While registration is tangentially related to the `-release` packages because of the implementation of `cloud-regionsrv-client` I still don't think that keg-recipes should have "knowledge" of this relationship. I think from a keg-recipes perspective we should break the `-release` packages into their own snippet.

But I do not think this is a blocker for this initial PR.

README.md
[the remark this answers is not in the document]
rjschwei: I'm not sure that is sufficient or meets our needs but let's see how it works out in practice. And if we end up with some duplication that's not the end of the world either.

data/base/cloudreg/sle15/packages.yaml
```diff
@@ -0,0 +1,15 @@
+packages:
+  bootstrap:
+    cloudreg:
+      - cloud-regionsrv-client
+      - regionsrv-certs
```
rjschwei, line 5: Revisit the inclusion of the `-release` packages here. I think they should be in a different snippet.

rjschwei (approved):

data/csp/ec2/ondemand/sle15/packages.yaml
jgleissner: Keys need to be unique (lowest level definition wins) so that's why the package lists are namespaced.

data/base/registration/sle15/packages.yaml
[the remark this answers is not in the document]
jgleissner: Just noticed, `cloud-regionsrv-client` and `regionsrv-certs` should not be in `bootstrap` but `image`. I'll fix it. I've put those together with the release files because they end up together in the images, so I didn't think further slicing would be necessary, but of course possible if a more logical separation is desired. Since I have to touch it anyway I can move them into separate snippets if you prefer.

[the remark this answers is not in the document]
rjschwei: Yes please. That way we can clearly show the difference between on-demand and byos in the `-release` packages we  add and keep it separated from the registration.

data/base/jeos/sle15/packages.yaml
schaefi: ok

data/csp/ec2/ondemand/sle15/packages.yaml
schaefi: :+1: no spaces anymore

schaefi (approved):
Looks good for an initial setup. Thanks

data/base/registration/sle15/packages.yaml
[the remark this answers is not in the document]
jgleissner: I've split it into `cloudreg` and `sle-modules`.

PR #2, opened 2020-12-22
Add data to generate config.sh

jgleissner:
This adds a template and SLE 15 configuration data to generate a config.sh file. The template isn't too pretty at the moment with some duplication. I intend to rewrite that into something nice, possibly moving some code into a shared template or maybe use auxiliary functions in keg. I'd just like to get agreement on the general approach first.


data/base/jeos/sle15/config.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - path: /etc/profile
+        append: True
+        content: |-
+          # yast in Public Cloud images fix
+          NCURSES_NO_UTF8_ACS=1
+          export NCURSES_NO_UTF8_ACS
+      - path: /etc/sysconfig/clock
+        append: True
+        content: |-
+          DEFAULT_TIMEZONE="Etc/UTC"
```
rjschwei, line 13: lower case 'e'

schemas/vm.config.sh.templ
```diff
@@ -0,0 +1,115 @@
...
+# FILE          : config.sh
+#----------------
+# PROJECT       : openSUSE KIWI Image System
+# COPYRIGHT     : (c) 2020 SUSE LLC. All rights reserved
+#               : 
+# AUTHOR        : {{ data['image']['author'] }} {{ data['image']['contact'] }}
+#               :
+# BELONGS TO    : Operating System images
+#               :
+# DESCRIPTION   : configuration script for SUSE based
```
rjschwei, line 37: Not limited to `SUSE`.

rjschwei (changes requested):

data/base/jeos/sle15/config.yaml
jgleissner: `Etc` is correct.

schemas/vm.config.sh.templ
jgleissner: Removed.

schaefi (commented):
Hi, thanks for the effort, this is hard work :+1: 

I have one design concern. The .yaml config parts contains raw data sections, for example the sed script that changes the PasswordAuthentication. The script code is placed under

```yaml
scripts:
    EC2-config: |-
        ...
```

My concerns are:

1. raw data in markup like yaml and json sucks. You can easily create strange effects depending on the data. bash scripting can contain all kinds of characters that can drive a yaml parser nuts or dumps data bash has problems with.

2. the scripting code is not dedicated to the yaml target. The setup of "no-ssh-password-auth" is something that is generically of interest and not only for EC2-config.

With regards to script creation I thought we can create a collection of generic shell functions and stack the script according to its functionality. In this particular case I would create a file
e.g

```bash
$ vi no-ssh-password-auth.sh

function no-ssh-password-auth {
    sed -i 's/#ChallengeResponseAuthentication yes/ChallengeResponseAuthentication no/' /etc/ssh/sshd_config
}
```

In the yaml I would then add:

```yaml
scripts:
    add-hook: no-ssh-password-auth
```

and some code with help of a template or not need to place the following in the final config.sh

```bash
source no-ssh-password-auth.sh

no-ssh-password-auth
```

I think it's not needed that the entire config.sh script is one big peace of code. Also if we create the yaml files by what it should do we could also share the tasks to actually implement the shell code to do that. Currently you are alone with all this. If we continue with the monolithic style it will be hard to share tasks between people.

What do you think ?

rjschwei:
Creating snippets as proposed by @schaefi LGTM

jgleissner:
Moving script code out of yaml is probably a good idea. There _shouldn't_ be any weird stuff in there that makes the yaml parser fall over, but I thought that about change logs too, so there's that. I moved all scriptlets into a 'scripts' directory in the data tree, in effect creating kind of a config.sh script library the data modules can use. I thought about having the scripts scattered in the data tree and loaded similarly to the yaml files, but that seems overly complicated. Unlike @schaefi suggestion, the snippets are simply expanded in config.sh (which was the easiest change in the schema). This could be changed into the functions approach, if there's a preference for that.

I've also updated the keg PR to support this.

Any thoughts?

schaefi (approved):
This is great now :+1: 

One last comment:

* The scripts are missing a shebang. I'd say we should add ```#!/bin/bash```. If we package this the suse checkers will complain otherwise and it's also better to indicate which shell interpreter is supposed to run the code

jgleissner:
It's intentional that there are no shebangs, as it's just script snippets that are assembled into a config.sh script. I believe the checker used during build only complains if scripts are executable so this shouldn't be an issue (should we ever feel the need to package keg recipes).

schaefi:
> It's intentional that there are no shebangs, as it's just script snippets that are assembled into a config.sh script. I believe the checker used during build only complains if scripts are executable so this shouldn't be an issue (should we ever feel the need to package keg recipes).

ok that makes perfect sense to me. Thanks

rjschwei (approved):

PR #4, opened 2021-03-02
Added JeOS Leap15 keg image definition

schaefi:
For demo and documentation purposes as well as to provide users
an easy start with Keg, I have added a Leap15 based image definition
which can also be build outside of obs and without access to SLES
repos

### DIFF
--- a/images/leap/jeos/15.2/image.yaml
+++ b/images/leap/jeos/15.2/image.yaml
@@ -1,0 +1,9 @@
+include-paths:
+  - base/jeos/leap15
+image:
+  name: Leap15.2-JeOS
+  specification: "Leap 15.2 guest image"
+  version: 1.15.2
+repositories:
+  -
+    path: obs://openSUSE:Leap:15.2/standard
--- a/images/leap/jeos/profiles.yaml
+++ b/images/leap/jeos/profiles.yaml
@@ -1,0 +1,5 @@
+profiles:
+  default:
+    description: Default Profile
+    include:
+      - csp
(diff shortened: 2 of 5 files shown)


schaefi:
Tested with

```
keg -r ~ms/Project/keg-recipes/ -d /tmp/mytest-desc/ leap/jeos/15.2 --force
sudo kiwi-ng --debug system build --description /tmp/mytest-desc/ --target-dir /tmp/mytest
```

images/leap/jeos/15.2/image.yaml
```diff
@@ -0,0 +1,9 @@
+include-paths:
+  - base/jeos/leap15
```
jgleissner, line 2: `include-paths` is intended for the version suffix not the whole thing. It should read (analogue to sles):

```yaml
include-paths:
  - leap15/1
  - leap15/2
```

images/leap/jeos/15.2/image.yaml
```diff
@@ -0,0 +1,9 @@
...
+image:
+  name: Leap15.2-JeOS
+  specification: "Leap 15.2 guest image"
+  version: 1.15.2
+repositories:
+  -
+    path: obs://openSUSE:Leap:15.2/standard
```
jgleissner, line 9: Will the default `obsrepositores` not work for leap?

images/leap/jeos/profiles.yaml
```diff
@@ -0,0 +1,5 @@
+profiles:
+  default:
+    description: Default Profile
+    include:
+      - csp
```
jgleissner, line 5: That doesn't do anything. There should be a profile that includes `base/jeos` as bare minimum, and CSP specific ones if you want to create images for public clouds.

PR #5, opened 2021-03-09
Overlay files

jgleissner:
This PR adds (most) of the files extracted from the various root overlay tarballs that exist at the moment. They are grouped functionally under `data/files/` and the idea is to include them in the configuration as shown in `data/csp/ec2/sle15/files.yaml`.

What do you think?


data/files/cloud-cfg-ali/etc/cloud/cloud.cfg
```diff
@@ -0,0 +1,69 @@
...
+ - phone-home
+ - final-message
+ - power-state-change
+
+# System and/or distro specific settings
+system_info:
+   distro: sles
+   # Default user name + that default users groups (if added/used)
+   default_user:
+     name: ec2-user
```
rjschwei, line 61: Wow, I still have not fixed that :( "ec2-user" should be "aliyun"

rjschwei (approved):
Generally the "named directory" approach works for me as it completely eliminates duplication and in the yaml that describes an image we just add then named directories from which to compose the overlay file.

I think there is an opportunity to understand the created differences for cloud-init configuration files by looking at these a bit more closely now that they are collected in one place.

jgleissner:
I've added configurations for the overlay files. It's not 100% complete but will hopefully be useful to go ahead with the implementation in keg (if all agree).

jgleissner:
A couple of things are sill missing from the overlay configuration, namely:

- /etc/iscsi/iscsid.conf

This is only included in the following images:

```
SLES12-SP2-SAP-EC2-HVM-BYOS
SLES12-SP2-SAP-EC2-HVM
SLES12-SP2-EC2-HVM-BYOS
SLES12-SP3-EC2-SAPCAL-HVM
SLES12-SP2-SAP-GCE-BYOS
SLES12-SP2-SAP-GCE
```

So this is almost obsolete, only SLES12-SP3-EC2-SAPCAL-HVM is still inactive. Given that it's only included in some images for some platforms, it would be a bit tricky to integrate nicely. Do we need to care about this?

- /opt/aws/bin/install-aws-agent.sh

Included in SLES 12 SAP images on EC2. Since it's only in EC2 and only in SAP it doesn't fit nicely in how the configuration is broken down. I would need to be in its own data module that is only included in the EC2 profile of the SAP image. As a side node, `SLES15-SP1-SAPCAL` includes the cloud.cfg directive to run this script but the script is not included. Either it needs to be added (if required on SAPCAP) or the cloud.cfg file should be dropped.

- /etc/systemd/system/sapinit.service.d/type.conf

Only in `SLES12-SP3-EC2-SAPCAL-HVM` and `SLES12-SP3-GCE-SAPCAL` and as such also doesn't fit in nicely. I wonder why it's not in the Azure image though?

rjschwei:
> A couple of things are sill missing from the overlay configuration, namely:
> 
> * /etc/iscsi/iscsid.conf
> 
> This is only included in the following images:
> 
> ```
> SLES12-SP2-SAP-EC2-HVM-BYOS
> SLES12-SP2-SAP-EC2-HVM
> SLES12-SP2-EC2-HVM-BYOS
> SLES12-SP3-EC2-SAPCAL-HVM
> SLES12-SP2-SAP-GCE-BYOS
> SLES12-SP2-SAP-GCE
> ```
> 
> So this is almost obsolete, only SLES12-SP3-EC2-SAPCAL-HVM is still inactive. Given that it's only included in some images for some platforms, it would be a bit tricky to integrate nicely. Do we need to care about this?

Well me may not care for this particular file but we care about the use case, which is "a specific file for multiple providers but only for certain images"


> 
> * /opt/aws/bin/install-aws-agent.sh
> 
> Included in SLES 12 SAP images on EC2. Since it's only in EC2 and only in SAP it doesn't fit nicely in how the configuration is broken down. I would need to be in its own data module that is only included in the EC2 profile of the SAP image. As a side node, `SLES15-SP1-SAPCAL` includes the cloud.cfg directive to run this script but the script is not included. Either it needs to be added (if required on SAPCAP) or the cloud.cfg file should be dropped.

Should be dropped from SAPCAL cloud.cfg. The agent is specifically to make things work nicely with the AWS maintained SAP qucik start. So the agent should be in all SAP images for EC2 and in all cloud.cfg files for SAP images in EC2.

> 
> * /etc/systemd/system/sapinit.service.d/type.conf
> 
> Only in `SLES12-SP3-EC2-SAPCAL-HVM` and `SLES12-SP3-GCE-SAPCAL` and as such also doesn't fit in nicely. I wonder why it's not in the Azure image though?

SLES 12 SP3 based SAPCAL image is on it's way out and is no longer getting updates, we can ignore this.

jgleissner:
Ok, so I'll simply remove modules `systemd-sapinit` and `iscsid-conf` since they are on the brink of obsolescence. As for the use case "specific to platform and product", as mentioned, this can be solved by adding a separate data module to be explicitly included in the appropriate profiles. AFAICS it's very rare (only the aws-agent at the moment really), so that approach should be serviceable IMO. I'll add a module for aws-agent.

jgleissner:
Overlay configuration adjusted as discussed. I set a couple of keys to 'null' to overwrite them, I thought it looks a bit nicer than using an empty include list. This will require a small change to keg though. I'd open a PR for that if people are ok with the approach.

rjschwei:
LGTM

schaefi (approved):
Looks good :+1:

PR #6, opened 2021-03-10
Fixed LI/VLI recipes

schaefi:
some value errors as well as an invalid type setup.
Also there is no oem config.sh template in the image
definition tree. I just copied the vm template for the
moment

### DIFF
--- a/data/csp/azure/li/profile.yaml
+++ b/data/csp/azure/li/profile.yaml
@@ -1,7 +1,7 @@
 profile:
   oemparameters:
     bootloader: grub2
-    bootpartition: false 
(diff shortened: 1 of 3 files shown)


jesusbv (approved):
Good catch

jesusbv:
> Also there is no oem config.sh template in the image
> definition tree. I just copied the vm template for the moment

Should we add a reminder/note to not forget adding/ changing it at a later time ?

schaefi:
I added https://github.com/SUSE-Enceladus/keg-recipes/issues/7 to track the missing oem templates

PR #8, opened 2021-03-10
Update scripts section to be more explicit

schaefi:
The current way of describing script data was only done
for one script (config.sh) and does not allow to distinguish.
We also need support for creating (images.sh) and thus this
commit updates the scripts such that they follow the
implementation from SUSE-Enceladus/keg#37


rjschwei (commented):
LGTM

PR #11, opened 2021-03-19
Azure LI/VLI refactor

jgleissner:
This moves the Azure LI and VLI modules into the own branch `csp/azure-baremetal`. There are a couple of things I'd like to get feedback on:

The package list contains a handful of SAP HANA specific packages. While we don't have another use case for Azure baremetal instances at the moment, theoretically you could have Azure baremetal instances where you don't want those. Those packages could be moved into another module, to be included in the `sles-sap/azure-{li,vli}` image definitions, to make it more granular. This could of course be done later if the need arises, but I thought I'd check for opinions.

The other thing is, the LI and VLI SLES 15 SP2 image descriptions as we use them at the moment differ in a way that doesn't seem to make sense (to me anyway), more specifically VLI includes a set of release packages LI doesn't:

```
sle-ha-release
sle-module-basesystem-release
sle-module-cap-tools-release
sle-module-containers-release
sle-module-desktop-applications-release
sle-module-development-tools-release
sle-module-legacy-release
sle-module-public-cloud-release
sle-module-sap-applications-release
sle-module-server-applications-release
sle-module-web-scripting-release
```

We generally don't install release files for modules in BYOS images so they should probably not be included in VLI builds either. I didn't include it in the recipes, but thought I'd double-check in case there is a reason for it.


jgleissner:
Any opinions?

schaefi:
sorry for the delay, looking at it today

schaefi:
> The package list contains a handful of SAP HANA specific packages. While we don't have another use case for Azure baremetal instances at the moment, theoretically you could have Azure baremetal instances where you don't want those. Those packages could be moved into another module, to be included in the sles-sap/azure-{li,vli} image definitions, to make it more granular. This could of course be done later if the need arises, but I thought I'd check for opinions.

I agree it would be good to have the HANA specific packages and settings separated. These settings are not specific to LI/VLI as the same HANA setup is used on the machines. So would be great to have that once and use it in LI/VLI ... and whatever will come next. I agree with you this can be done in a followup PR

schaefi:
> The other thing is, the LI and VLI SLES 15 SP2 image descriptions as we use them at the moment differ in a way that doesn't seem to make sense (to me anyway), more specifically VLI includes a set of release packages LI doesn't:

This is interesting. I wasn't aware of that difference. I agree they should not be installed. Thanks for catching this

schaefi:
If I read the PR correctly the release issue is already covered. So this looks good to me

schaefi (approved):

rjschwei (approved):

rjschwei:
Sorry for the slow review

PR #12, opened 2021-04-23
Add support for archive tags

jgleissner:
This PR adds archive tags to the schema templates. There were already archive tags referenced in the templates, but those don't exist in that form in the data dictionary and hence were effectively no-ops.


jgleissner:
This PR requires https://github.com/SUSE-Enceladus/keg/pull/47 to be functional.

rjschwei (approved):

PR #13, opened 2021-04-27
Split EC2 data module

jgleissner:
Moved packages from base EC2 module into csp/ec2/addon/aws-tool so CHOST canuse a stripped down set of EC2 packages.
Moved ECS module into csp/ec2/addon.
Added aws-sap-agent in csp/ec2/addon (will be used in SLE12 images).

### DIFF
--- a/data/csp/ec2/addon/aws-sap-agent/sle12/overlayfiles.yaml
+++ b/data/csp/ec2/addon/aws-sap-agent/sle12/overlayfiles.yaml
@@ -1,0 +1,4 @@
+overlayfiles:
+  ec2-addon-aws-sap-agent:
+    include:
+      - script-install-aws-agent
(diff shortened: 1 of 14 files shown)


jgleissner:
Btw, I noticed that some SLES 12 images include `python-ec2metadata` and some `python3-ec2metadata`. Is this on purpose or should all use the python3 version?

rjschwei:
@jgleissner `python3-ec2metadata` did not get pushed all the way back to SLES 12, but I think started inSLES 12 SP1 repos. By the time we switched to `python3-ec2metadata` 12 SP1 may have been out of support already as such the difference. This should remain as is for historical purposes/re-construction if necessary.

rjschwei (approved):

jgleissner:
> @jgleissner `python3-ec2metadata` did not get pushed all the way back to SLES 12, but I think started inSLES 12 SP1 repos. By the time we switched to `python3-ec2metadata` 12 SP1 may have been out of support already as such the difference. This should remain as is for historical purposes/re-construction if necessary.

That's fine. There are SLES12 SP4 images that use `python-ec2metadata` as well though. Perhaps they simply weren't updated because they are inactive already, so I guess using `python3-ec2metadata` in keg-recipes for those shouldn't be an issue.

PR #14, opened 2021-05-05
Merge kiwi template into one-size-fits-all version

jgleissner:
Drop oem and vm_singlebuild version and absorb functionality in default vm.kiwi.templ.
Move dict_to_param_string macro into macros.jinja.

This does away with duplication in the template files. The `oem` version was mostly added for the Azure LI/VLI images, and their parameter structure is a bit different from the VM images, and it seemed easier to deal with this moving the image parameters into another namespace, but since Azure baremetal resides in its own space now, this isn't really necessary.

The idea of `vm_singlebuild` was to have a simpler template, but it turned out to be barely simpler and supporting both in one template doesn't increase complexity much. IMO not having duplication in the templates outweighs that significantly. Also, since the single and multi build data is treated differently, we have some special sauce in `keg` to handle that, which could be dropped with this approach. Single build is automatically used when there is only a 'common' profile defined. There's a slight drawback that image definitions that inherit multi build profiles would need to overwrite them to become single build, but I think this would be acceptable.


rjschwei (approved):
LGTM

PR #15, opened 2021-05-06
Refactor SLE15 data

jgleissner:
Sorry, this PR is way too big.

Changes overview
-------------------------

Split base/sles into base/sle and products/sles.
Move base/chost to products/chost.
Move base/hpc to products/hpc.
Add Aliyun module.
Break some CSP specific packages off into 'addon' modules. This allows for
leaner images (e.g. CHOST).
Add kernel-azure addon module.
Use default template for Azure LI/VLI images.
Use lowercase for namespaces for consistency.
Fine-tune package selections.
Move config:config_script: to config: and config:image_script: to setup:.

Details
---------

Splitting `base/sles` into `base/sle` and `products/sles` has two purposes. It creates a common SLE base for products to use, but moves some packages out that are not in all the images. Also, `sles-release` conflicts with other product release packages so it must only be included in a SLES image.

Moving chost and hpc to `products` was done because including those modules makes the image a 'product' (i.e. inlcudes a product release package).

Breaking off CSP specific packages into separate modules was basically done so they wouldn't need to be included in CHOST.

I changed the Azure LI and VLI images to use the unified template of https://github.com/SUSE-Enceladus/keg-recipes/pull/14. This also applies to the EC2-ECS image, which I changed to singlebuild. Forgot that in the commit log.

I moved `config:config_script:` and `config:image_script:` keys to root as `config:` and `setup:`. The additional hierarchy level didn't really serve a purpose. I've adjusted the templates, although I'm planning to get rid of those as per @schaefi's suggestion, as using a template engine for piecing the scripts together doesn't really provide a benefit.

The namespace keys' naming wasn't consistent (mixed lower- and uppercase). I changed it to use lowercase everywhere (except for the profile names). Please complain if you prefer uppercase.

A lot of fine-tuning and moving around package lists. This makes the PR especially bloated, sorry about that.

Open questions
---------------------

I've set up test builds and compared the package lists with the ones from the existing builds and they match pretty well, with a few exceptions.

- system-group-wheel

This is in some images, missing in most. Any objections to include this in all images?

- python3-google-auth

This package and a few related ones are in GCE SLES for SAP but not in SLES. Should they really be in SLES for SAP only?

- SuSEfirewall2 (in Azure LI/VLI)

This package is obsolete and unmaintained, so it should not be included. Is it ok to replace it with `firewalld`?

- salt-minion (not in Azure LI/VLI)

This is in all BYOS images except Azure LI/VLI. Should it be included?

- sap-suse-cluster-connector
- system-role-sles4sap
- socat

Those are in SLES for SAP images but not Azure LI/VLI. Should we include them?

- audit
- kdump
- terminfo

Those are not in the released EC2-ECS image but since those packages are in everything else they are in `base/jeos` at the moment and hence get  included in the keg generated version. Is it ok to include them going forward?

- cloud-netconfig (not in CHOST)

The released (and kegged) CHOST images don't contain cloud-netconfig. Wouldn't it be useful though to include it? It's a few shell scripts so the package is really small.

- docker in Manager images

The released SUSE Manger images contain docker, not sure whether that makes sense. Should we leave it in or drop it?


rjschwei:
Handling questions separately form the review and first:

> system-group-wheel
> This is in some images, missing in most. Any objections to include this in all images?

No can and should be in all images.

> python3-google-auth
> This package and a few related ones are in GCE SLES for SAP but not in SLES. Should they really be in SLES for SAP only?

Yes SLES4SAP only, is needed by the resource agent.

> SuSEfirewall2 (in Azure LI/VLI)
> This package is obsolete and unmaintained, so it should not be included. Is it ok to replace it with firewalld?

Yes should be replaced by firewalld

> salt-minion (not in Azure LI/VLI)
> This is in all BYOS images except Azure LI/VLI. Should it be included?

salt-mnion is in the BYOS images to make setup automation via terraform templates possible, i.e. enable the SHAP team automation. There is no automation in LI/VLI as such having the package in those images would be superfluous. Anyone wanting to manage LI/VLI with SUMa will need to enroll those systems and then they get salt-minion as part of the bootstrap process.

> sap-suse-cluster-connector
> system-role-sles4sap
> socat
> Those are in SLES for SAP images but not Azure LI/VLI. Should we include them?

Those should not be missing in LI/VLI they are all important to set up HA

> audit
> kdump
> terminfo
> Those are not in the released EC2-ECS image but since those packages are in everything else they are in base/jeos at the moment and hence get included in the keg generated version. Is it ok to include them going forward?

Yes OK to include for now. When we have SLE-Micro sorted out the ECS image will be re-based, or it will switch to a CHOST like implementation.

> cloud-netconfig (not in CHOST)
> The released (and kegged) CHOST images don't contain cloud-netconfig. Wouldn't it be useful though to include it? It's a few shell scripts so the package is really small.

Well I do not know if the stack above, i.e. the VLAN setup code in docker/podman/systemd properly handles multiple interfaces. I am not certain to container world is at that point just yet. I'd say we leave it as is until we get a request to support multiple interfaces on container hosts.

> docker in Manager images
> The released SUSE Manger images contain docker, not sure whether that makes sense. Should we leave it in or drop it?

docker was added at some point everywhere to make it easier for customers to run containers. Of course since then podman has emerged and docker is somewhat on it's way out. I don't necessarily see a reason to remove docker from the SUMa images, but am also fine to drop it. SUMa itself will be containerized at some point and as such we'll get a container setup engine back in some way shape or form.

rjschwei (approved):
LGTM

The only thing I'd say is that we might want to rethink the name use for `jeos`. We might eventually want the "jeos" team to move their image descriptions into keg-receipies as well and our `jeos` is just a bit different. Avoiding this name conflict now may be easier than trying to do it later.

jgleissner:
> The only thing I'd say is that we might want to rethink the name use for `jeos`. We might eventually want the "jeos" team to move their image descriptions into keg-receipies as well and our `jeos` is just a bit different. Avoiding this name conflict now may be easier than trying to do it later.

Good point. The `jeos` module in keg is really the set of packages that is in everything. Probably a good idea to avoid ambiguity with a potential JeOS product. Maybe call it `minimal`? Or `common`?

rjschwei:
`minimal` also has a meaning in SLE already as in the `minimal pattern`. `common` works for me

jgleissner:
> > python3-google-auth
> > This package and a few related ones are in GCE SLES for SAP but not in SLES. Should they really be in SLES for SAP only?
> Yes SLES4SAP only, is needed by the resource agent.

Ok, will put them in an `addon` module for GCE.

> > docker in Manager images
> > The released SUSE Manger images contain docker, not sure whether that makes sense. Should we leave it in or drop it?
> 
> docker was added at some point everywhere to make it easier for customers to run containers. Of course since then podman has emerged and docker is somewhat on it's way out. I don't necessarily see a reason to remove docker from the SUMa images, but am also fine to drop it. SUMa itself will be containerized at some point and as such we'll get a container setup engine back in some way shape or form.

Ok. In keg-recipes `docker` is currently not in proxy but still in server. I guess running docker workloads on SUMa hosts isn't really recommended so maybe we shouldn't include it, but I don't have a strong opinion on this.

jgleissner:
I've added the `google-auth` module and renamed `base/jeos` to `base/common`, and also pushed a couple of tweaks. From my point of view the PR is good to be merged.

PR #17, opened 2021-05-18
Replace script templates with header only templates

jgleissner:
This PR drops the full blown script templates and only leaves small templates to generate script headers. This corresponds to https://github.com/SUSE-Enceladus/keg/pull/50.


schemas/config_sh_header.templ
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#================
+# FILE          : config.sh
+#----------------
+# PROJECT       : SUSE Public Cloud recipes
+# COPYRIGHT     : (c) 2021 SUSE LLC. All rights reserved
```
rjschwei, line 6: The year should be a variable so we do not have to touch the template every year to update.

schemas/images_sh_header.templ
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+#================
+# FILE          : image.sh
+#----------------
+# PROJECT       : SUSE Public Cloud recipes
+# COPYRIGHT     : (c) 2021 SUSE LLC. All rights reserved
```
rjschwei, line 6: ditto copyright year as variable

rjschwei (changes requested):

jgleissner:
It's certainly possible to generate the year in the copyright notice. Do we actually need one in a generated script?

rjschwei:
> It's certainly possible to generate the year in the copyright notice. Do we actually need one in a generated script?

Well if we call out the copyright notice, which I think we are expected to, then that notice does have to have a year. Whether the code is generated or manually created is immaterial.

rjschwei (approved):

PR #19, opened 2021-06-07
Delete use of deprecated kiwi config function

schaefi:
suseImportBuildKey is deprecated and done by zypper
via the --gpg-auto-import-keys option used when kiwi
calls the package manager

### DIFF
--- a/schemas/config.sh.templ
+++ b/schemas/config.sh.templ
@@ -66,11 +66,6 @@ set -e
 #--------------------------------------
 echo "Configure image: [$kiwi_iname]..."
 
-#======================================
-# Setup the build keys
-#--------------------------------------
-suseImportBuildKey
-
 {% set common_profile = data['profiles']['common'] %}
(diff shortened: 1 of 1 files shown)


schaefi:
In kiwi upstream the use of deprecated methods will lead to a build error. This change is done to avoid any issue when this change made its way to SLE.

The use of the method at the moment is also not doing anything useful unless the gpg pubkey packages for all used repos are installed during the bootstrap phase. Thus the use of this method could have been dropped some time ago.

The same change was applied to the Azure LI/VLI images from where you can see that it did not harm the build
e.g Devel:PubCloud:CrossCloud:SLE15-SP3/SLES15-SP3-SAP-Azure-LI-BYOS

Thanks

jgleissner:
Looks good to me but shouldn't it go into branch `develop` first?

schaefi:
yes sorry

PR #26, opened 2021-06-09
SAP: Move libvirt-client into addon/docker module

jgleissner:
This prevents libvirt-client from being added to Azure baremetal image descriptions.


rjschwei (approved):
LGTM

PR #29, opened 2021-07-01
SLE15 updates

jgleissner:
This PR adds SLE15 SP1 image definitions including a few minor alterations to the existing data to accommodate for differences in released SP1 images compared to SP2. I'm still analyzing image binaries so a few changes may still come, but should be minor.

There are a couple of things I'd like to have feedback on:

- SAPCAL module

I've added the SAPCAL module under `data/products/sapcal` even though it's not really a product but really just a SLES flavor. The module does not install a `release` package like the other modules in `data/products`. I thought about adding it under `data/base` instead but it feels more like a product module rather than a base module. I don't really have a strong opinion either way. If you think it shouldn't be a product module I'm happy to move it.

- ResourceDisk.Format setting in /etc/waagent.conf

In Azure SP1 BYOS, SP1 SAP BYOS, and SP2 BYOS `ResourceDisk.Format=n` is set in `/etc/waagent.conf`. I haven't added this yet to the keg recipes. Is this really only supposed to be in those images?

- firewalld

SLES 15 SP1 BYOS images include `firewalld`, which was apparently removed from SP2 images. I've added `firewalld` to `base/sle/sle15/sp1` just to remove it again in `base/sle/sle15/sp2`. AFAICS no on-demand SP1 image that didn't include `firewalld` are still maintained, so this simplified approach should be fine IMO. Please object if it's not.

- zypper-lifecycle-plugin

This was only added after SP1. I didn't bother removing it from SP1 images in keg, but can do if desired.

- grub timeout style in Azure images

Azure SP2+ sets grub timeout style to `countdown`, it wasn't set before. I didn't bother removing it in the keg version, but can do if it's somehow important.

- SAP specific packages

SLES 15 SP2 SAP contains a number of packages that are not included in SP1, namely:

```
drbd-formula
HANA-Firewall
habootstrap-formula
salt-master
salt-shaptools
salt-standalone-formulas-configuration
sap-suse-cluster-connector
saphanabootstrap-formula
sapnwbootstrap-formula
saptune
socat
```

Is this is on purpose? Should I remove them from the SP1 keg versions as well?

Thanks


rjschwei:
@jgleissner I'd rather treat `sapcal` as a flavor along the lines of `chost` and `ec2-ecs` for example.

rjschwei:
@jgleissner the `ResourceDisk.Format=n` in `/etc/waagent.conf` was added in the build service only to images that have the potential to be refreshed per life-cycle to reduce work with cut and paste. Nominally yes the setting should be in all images. However given that we have never released images with that setting images thta wil not get refreshed should not get that setting in the repo to avoid the impression that we ever had an image with that setting the PAYG and earlier than 15 SP3.

rjschwei:
@jgleissner Sorry I do not follow your train of thought w.r.t. `firewalld`

rjschwei:
@jgleissner I only added `zypper-lifecycle-plugin` to 15 SP3 images. The original request was to have it in 15 SP2 images. However since that bug didn't get addressed until after the release of 15 SP3 it makes no sense to add it to the description for images that will no longer get refreshed.

I think this is a general guideline we should follow. For images that no longer get refreshed we should produce the final configuration that exists in the build service. We do not want to give the impression that somehow those images were released with packages/settings that are actually not in the image. Since this can get horribly complicated because of the inconsistencies inherent to the manual maintenance of many image descriptions we may also choose not to publish descriptions in keg-recipes for images that no longer get refreshed.

rjschwei:
@jgleissner Since SLES4SAP images still get refreshed, potentially, I think adding all the packages added to 15 SP2 back to 15 SP1 and 15 is OK. But we have to also update the descriptions as they exists in the build service today to make sure we pick these up on the next refresh.

rjschwei:
Since based on the comments there is some re-arrangement needed, I think I will skip the review for now until the re-arrangement is complete.

jgleissner:
> @jgleissner I'd rather treat `sapcal` as a flavor along the lines of `chost` and `ec2-ecs` for example.

The image definition is located in `image/sles/sapcal`, reflecting its status as a SLES flavor, analogue to `chost` and `ec2-ecs`. The question is about the data module, which I've put into `data/products`, even though it doesn't, unlike all the other modules in that directory, install a product release package like `sles-release`. As said, I didn't put it in `data/base` because it doesn't really "feel" like a base module, but I can move it there, or we can invent another module category.

jgleissner:
Wrt to changing image descriptions of released and especially inactive image, I agree that this should be avoided, of course. I'll change the recipes to include `zypper-lifecycle-plugin` only from SP3. I can make the additional SAP  packages SP2+ only too, but in case we would add them anyway if we ever needed to refresh the inactive images, it wouldn't make sense, hence the question.

jgleissner:
Regarding `firewalld`, my point was that adding it to `data/base/sle/sle15/sp1` enables it in most SLES 15 SP1 images, even though it wasn't released with SLES 15 SP1 on-demand images. Since we don't maintain those anymore and hence won't need to generate descriptions for them, I think it's ok.

jgleissner:
Are we ok to merge this? I noticed that there are inactive SLES 15 GA images too, but I suggest adding them in a separate PR.

rjschwei (approved):

PR #32, opened 2021-08-04
SLE 12 data and image definitions

jgleissner:
This PR does the following:

- Add a kiwi7 compatible schema for SLE 12 SP3 image descriptions (the `users` section as used in the newer descriptions isn't supported by kiwi7)
- Move SLE 15 profile.yaml files into sle15 sub directories to completely separate SLE 12 and SLE 15 data
- Add SLE 12 data modules plus required overlay files and scriptlets
- Add SLE 12 image definitions under `images/sle-legacy` (except some Azure LI/VLI definitions)
- Move SLE 15 SP1 image definitions into `images/sle-legacy`
- Minor fixes to SLE 15 data

I ended up using a separate directory `images/sle-legacy` for the image definitions that do not use the profile scheme of SLE 15 SP2+ image definitions. Trying to shoehorn those in turned out to not work well.


rjschwei (approved):
It's really too much for a detailed review. Nothing stood out in a quick look. Everything else can be fixed with subsequent commits

jgleissner:
Sorry, this just kept growing. There wasn't really a particularly good way to break it down into smaller bits because it's all intertwined, so trying to do it in installments would have meant changing data all the time. There are a couple of minor things still open, but I'll just do separate PRs for them.

rjschwei:
On ething I missed here, rather than `sle-legacy` we should use something like `single-build` or `no-profiles`

jgleissner:
> One thing I missed here, rather than `sle-legacy` we should use something like `single-build` or `no-profiles`

I thought of `single-build` but there is `images/sle-legacy/12-sp5` which is multi-build so I opted for `sle-legacy` as kind of a bucket term. I meant `legacy` in terms of image description layout (i.e. before we adopted the current one with `15-sp2`) rather than referring to the images, but I see how it could be misleading. I don't have a better idea though.

rjschwei:
Given that we are sensitive to this I think we should come up with descriptive names, `legacy` is just associated with to many "bad" connotations. I suggest we move everything that matches the current layout into `cross-cloud` and use `single-cloud` as entry point for the descriptions where we built an image for a given framework in on description.

jgleissner:
`single-cloud` works, good idea. Not sure whether we really need `cross-cloud` as an additional hierarchy level but I don't have a strong opinion on this. I have some local commits within the existing structure so I'd do PRs for those first and move the image definitions afterwards in a separate PR.

rjschwei:
I think it will serve us well not to treat our current  thinking as the holy grail and as such having `cross-cloud` and `single-cloud` makes it more obvious that these are equally valid organizational approaches. Creating `cross-cloud` as an entry point is a "do when convenient" operation as far as I am concerned.

Issue #38, opened 2021-08-25
Handling of rpm-ndb


aosthof:
Some time ago we had a request to replace rpm by rpm-ndb because of the Sleepycat license used by the Berkely-DB which could be problematic for some customers. According to PM-2071 we started with all SLES 15 SP2 based images. However this change possibly went also in older distributions.

We need to include the following line in our config.kiwi files in order to make IBS use rpm-ndb instead of rpm:

`<!-- OBS-IgnorePackage: rpm -->`

jgleissner:
We can get rid of the OBS specific comments in the template, of course. We'll still have OBS specific knowledge in keg-recipes, obviously, but having the data in the image definitions rather than in the template may seem less intrusive to people who want to use the generated descriptions without OBS.

My question was more aimed at the location we should support configuration of these comments. Directly in the image definition (i.e. somewhere in the `images/` tree) feels natural but it means that you potentially have duplication, at least with the `OBS-IgnorePackage` comment, since that was only introduced with SP2, and there are a few older image definitions in `images/cross-cloud`, so it'd be either adding that comment to all SP2+ `image.yaml` files or let them all inherit it and disable it in <SP2 ones. Which might be acceptable.

jgleissner:
Forgot to close this. This was fixed with the merge mentioned above.

PR #40, opened 2021-09-02
Configurable comments in kiwi schema

jgleissner:
Implement support for adding comments to generated kiwi schema.
Change all existing arch and build flavor config to schema comments.
Add OBS-IgnorePackage: rpm comment (https://github.com/SUSE-Enceladus/keg-recipes/issues/38).


rjschwei (approved):
`schema-comments` should probably be `image-config-comments` since keg does not generate a schema.

jgleissner:
> `schema-comments` should probably be `image-config-comments` since keg does not generate a schema.

I was unsure about the name. What we include now is really build configuration (why OBS needs this to be in `config.kiwi` is perhaps another topic), so not actually image config data. But then the comments could be anything, even just informal ones. So I settled on `schema-comments` because it's comments that get injected into the schema. I'm open to better a better name, of course.

rjschwei:
> > `schema-comments` should probably be `image-config-comments` since keg does not generate a schema.
> 
> I was unsure about the name. What we include now is really build configuration (why OBS needs this to be in `config.kiwi` is perhaps another topic), so not actually image config data. But then the comments could be anything, even just informal ones. So I settled on `schema-comments` because it's comments that get injected into the schema. I'm open to better a better name, of course.

But `kiwi.config` is not a `schema` that's what I think makes things a bit weird looking/reading. And `image-config-comments` is generic enough to express that it could be anything. But I am alos fine leaving he name. I'll get over it.

jgleissner:
Initially was using `kiwi-comments`, as it appears in the kiwi file (I don't think there's an established term for config.kiwi except "the kiwi file"). I can change it to `image-config-comments`, no problem.

PR #42, opened 2021-09-29
Add runc 

rjschwei:
- Add runc to CHOST images bsc#1190827


jgleissner:
This will also add runc to the 15 SP1 version. Is that ok? The bug report only mentions SP2+.

rjschwei:
> This will also add runc to the 15 SP1 version. Is that ok? The bug report only mentions SP2+.

Yes that is OK. The bug report mentions SP2 because that is what SAP is currently using.

jgleissner (approved):

PR #47, opened 2021-10-07
Fix Azure VLI kernel cmd line

jgleissner:
Remove 'tsc=reliable' from kernelcmdline in sle12 profile and add it in sle15 profile.


schaefi:
looks good to me :+1:

PR #51, opened 2021-10-12
Do not set boottimeout in SLE15 Azure baremetal

jgleissner:
https://github.com/SUSE-Enceladus/keg-recipes/pull/49 sets a default "old style" `boottimeout` paramater for SLE15 <=SP1 that is unwanted in Azure LI/VLI. This PR clears the parameter in the azure-baremetal module.


jgleissner:
Can we merge this?

PR #53, opened 2021-10-21
Minor fixes Azure LI/VLI

jgleissner:
Drop all salt related packages.
Drop fio.
Drop libvirt-client.
Add default timout and nmi_watchdog kernel parameters.


schaefi (approved):
Looks good to me :+1: Thanks

PR #63, opened 2021-10-28
Set rpm-excludedocs in CHOST SLE15 SP2+

jgleissner:
Adds support for configurable global preferences tags and uses that feature to set `rpm-excludedocs` for CHOST images (SLE15 SP2+).


jgleissner:
Ok to merge?

PR #69, opened 2021-11-03
- Add apparmor-abstractions to CHOST for 15 Sp2 and 15 SP3 build

rjschwei:
+ This addresses bsc#1190829
  + Containre termination with runc is broken without this package

### DIFF
--- a/data/products/chost/sle15/sp2/packages.yaml
+++ b/data/products/chost/sle15/sp2/packages.yaml
@@ -1,6 +1,7 @@
 packages:
   image:
-    chost:
+    chost-apparmor:
       - apparmor-abstractions
+      - apparmor-parser
     chost-btrfsprogs: Null
     chost-fcoe-utils: Null
--- a/data/products/chost/sle15/sp2/packages.yaml
+++ b/data/products/chost/sle15/sp2/packages.yaml
@@ -1,4 +1,6 @@
 packages:
   image:
+    chost:
+      - apparmor-abstractions
     chost-btrfsprogs: Null
     chost-fcoe-utils: Null


data/products/chost/sle15/sp2/packages.yaml
```diff
@@ -1,4 +1,6 @@
 packages:
   image:
+    chost:
```
jgleissner, line 3: This would override the `chost` namespace from `products/chost/sle15/packages.yaml` and thus remove the base CHOST package selection. Assuming this is not intended, I suggest renaming the namespace to something like `chost-apparmor-abstractions`.

jgleissner (changes requested):

data/products/chost/sle15/sp2/packages.yaml
rjschwei: But hen that new namespace has to be included somewhere, right? Meaning somewhere along the line the rest of the code has to know that `chost-apparmor-abstractions` has to be picked up. But that was not the intention. The idea was always that one can add some packages to a given base image by addition, meaning in this case the `chost` package specification needs to be merged with wherever else `chost` may be specified. Remember the "modified depth first search" approach we discussed. I am fraid this turns things into a big problem.

jgleissner: Namespaces are not included by the image definition directly. Image definitions work on path level. E.g. the `common` profile definition for chost in `images/cross-cloud/sles/chost/profiles.yaml` just includes `products/chost`, and the image definition adds the version specific sub paths. All namespaces in those paths are merged together. The namespaces serve the purpose of grouping data and allow for an override mechanism in more specific layers.

Or in short: If you  drop a new namespace in `data/products/chost/sle15/sp2`, you extend the chost base image for 15 SP2+ (no further changes required), which I think is exactly what you want to do.

rjschwei: Thanks for the explanation. Fixed.

jgleissner (approved):

Issue #75, opened 2021-11-22
Handle kiwi schemaversion


jgleissner:
Currently, the `schemaversion` is hard-coded to version 6.2 in the jinja template. Some configuration snippets in keg-recipes however require a later schemaversion, e.g. the switch to separate bootloader section implemented in `data/csp/sle15/sp2/profile_defaults.yaml`. So some configuration option would be required to express the minimum schemaversion required.

A straight-forward solution would be to add `schemaversion` to the image definition in `data/images`. This would mean that the image definition declares the schemaversion it complies with, even though it's really the configuration bits in the data modules that add define the requirement. Adding that information directly where the schemaversion requirements are created seems more logical, but because this is data that is merged into the profile specific sections, you can't really specify global parameters at that point. This would require support in keg to be aware of that schemaversion specific settings to figure out which schemaversion to use (i.e. the highest one that is set in any included modules).

However, this may be a bit more complicated than necessary, so I'd like to get some opinions on this. @rjschwei @schaefi, any input?

schaefi:
I think it would be sufficient to add one schema version in data/images and then care for the single config snippets to follow this schema. This means it's a responsibility of the keg recipes author to create correct data and for the people to review changes.

In the end if it's done wrong it will be "fixed" through the kiwi schema templates or fail at that level if this is not possible.

If we allow multiple schema version constraints in several snippets that gets pulled together we need some mechanism in keg itself to make sure only one is taken as the final schema version setting. The advantage here is that we can prevent merging configuration elements of different schema versions.

Thinking about the two, the first process will lead more often to a good result and will safely fail in case of an error.
If we find out it will be better to specify the schema version per config snippets it can still be done on top of the schema version configured in data/images

Thoughts ?

jgleissner:
I'm also leaning towards approach no 1, a.k.a. the simple one. Allowing to define which schema version a configuration snippet applies to might be neat, but figuring out which schema version the final description applies to is not necessarily straight-forward. Some options may be comply with later schema versions, other don't, so you'd need to configure that as well. And it would make things more complicated on keg side, perhaps with little practical benefit. Plus, if it turns out that we do want to have a more sophisticated handling of schemaversion down the line, we can always implement it.

jgleissner:
Fix merged.

PR #77, opened 2021-11-26
Set appropriate kiwi 'schemaversion' in all images

jgleissner:
Make 'schemaversion' an image config option.
Set '6.2' as default 'schemaversion'.
Set 'schemaversion: 7.1' in images/cross-cloud/sles/chost/15-sp1.
Set 'schemaversion: 7.2' in all SLE15 SP2+ image definitions.

This implements the 'simple' approach as outlined in https://github.com/SUSE-Enceladus/keg-recipes/issues/75. As mentioned, it may be nicer to set the schema version where the requirement for a specific version is created, e.g. in `data/csp/sle15/sp2` which changes the boot config style to schema 7.2. This would also reduce duplication as `schemaversion` is now in every image definition that doesn't use the default. However, it would make things more complicated in keg so I went the easier route, for now anyway.

### DIFF
--- a/images/cross-cloud/sle-hpc/byos/15-sp2/image.yaml
+++ b/images/cross-cloud/sle-hpc/byos/15-sp2/image.yaml
@@ -1,6 +1,7 @@
 include-paths:
   - sle15/sp1
   - sle15/sp2
+schemaversion: 7.2
 image:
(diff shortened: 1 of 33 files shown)


rjschwei (approved):

aosthof (approved):
LGTM, thanks!

PR #80, opened 2021-12-02
- Include the registration client in all images to support BYOS instance

rjschwei:
using the update infrastructure


rjschwei:
@jgleissner this has no effect on config.sh generation correct? For BYOS we do not want to enable the regionsrv client service by default.

jgleissner:
The guestregister service gets enabled by `data/base/cloudreg/sle?/config.yaml`. Since we're getting rid of `data/base/cloudreq`, those configs would need to be replicated in `data/csp/*/ondemand/sle?`.

images/cross-cloud/sle-hpc/ondemand/profiles.yaml
```diff
@@ -10,17 +9,14 @@ profiles:
...
   EC2:
     description: EC2 configuration
     include:
       - csp/ec2/addon/aws-tools
-      - csp/ec2/ondemand
   GCE:
     description: GCE configuration
     include:
       - csp/gce/addon/gce-tools
-      - csp/gce/ondemand/hpc
```
jgleissner, line 26: We'll have to retain the `ondemand` includes. Firstly, because we probably want to put the config section that enables `guestregister.service` there, and secondly because we use those paths to slot in the billing tags from the private repo.

jgleissner (changes requested):

data/csp/gce/ondemand/config.yaml
```diff
@@ -0,0 +1,4 @@
+config:
+  services:
+    cloudreg:
+      - guestregister
```
jgleissner, line 4: While this would work, we decided at some point to completely separate SLE 12 and 15 configurations. It creates duplication but keeps it simpler overall and avoids accidentally changing something in the other code stream. So this should be in `sle12` and `sle15` sub dirs under `ondemand`.

data/csp/gce/ondemand/overlayfiles.yaml
```diff
@@ -0,0 +1,4 @@
+overlayfiles:
+  motd-ondemand:
+    include:
+      - motd-addon-mgmt
```
jgleissner, line 4: Ditto.

images/cross-cloud/sle-hpc/ondemand/profiles.yaml
```diff
@@ -10,10 +9,11 @@ profiles:
   Azure:
     description: Azure configuration
     include:
-      - csp/azure/ondemand/hpc
       - csp/azure/addon/azure-kernel
       - csp/azure/addon/azure-tools
       - csp/azure/addon/waagent-rdma
+      - csp/gce/ondemand
```
jgleissner, line 15: This probably meant to read `azure` instead of `gce` (in which case it would be superfluous). Should be removed.

images/cross-cloud/sle-hpc/ondemand/profiles.yaml
```diff
@@ -23,4 +23,5 @@ profiles:
     description: GCE configuration
     include:
       - csp/gce/addon/gce-tools
+      - csp/gce/ondemand
```
jgleissner, line 26: This line isn't necessary because the line `csp/gce/ondemand/hpc` below already includes all parent directories. It doesn't harm either, and may be used to create more descriptive image definitions I suppose. So far I've always opted for the most compact definition though.

jgleissner (changes requested):

rjschwei:
Now that we have to make modifications I am wondering about the separation by convention. But that is for another time. So here is my next attempt.

images/cross-cloud/sles-sap/ondemand/profiles.yaml
```diff
@@ -1,14 +1,14 @@
...
     include:
-      - base/cloudreg
       - base/common
       - base/sle
       - base/sle-modules
       - products/sap/ondemand
   Azure:
     include:
       - csp/azure/addon/azure-tools
+      - csp/azure/ondemand
```
jgleissner, line 11: This is superfluous as `csp/gce/ondemand/sap` already includes it. Unless you want it for cosmetic reasons, I suggest removing it.

images/cross-cloud/sles-sap/ondemand/profiles.yaml
```diff
@@ -19,4 +19,5 @@ profiles:
       - csp/gce/addon/fence-agent-deps
       - csp/gce/addon/gce-tools
       - csp/gce/addon/google-auth
+      - csp/gce/ondemand
```
jgleissner, line 22: Superfluous.

images/cross-cloud/sles/ondemand/profiles.yaml
```diff
@@ -28,4 +27,5 @@ profiles:
     description: GCE configuration
     include:
       - csp/gce/addon/gce-tools
+      - csp/gce/ondemand
```
jgleissner, line 30: Superfluous.

jgleissner (changes requested):
I suggest removing the superfluous lines in the image descriptions, but otherwise looks good.

jgleissner (approved):

PR #90, opened 2022-01-14
Disable cgroup kernel memory tracking for CHOST images 15 SP2 and later

rjschwei:


jgleissner (approved):
Technically, parameters `cgroup_enable` and `swapaccount` are superfluous, but don't do harm.

PR #95, opened 2022-02-02
Add iscsi-formula to SLES for SAP 15 SP4

jgleissner:


jgleissner:
This will also add it to Azure LI/VLI and may drag salt into the build, which is probably undesirable.

jgleissner:
I'll open a new PR, forgot to include the bsc# anyway.

PR #96, opened 2022-02-02
Add iscsi-formula to SLES for SAP 15 SP4 (bsc#1195225)

jgleissner:


jgleissner:
Still not quite right, new PR coming.

PR #103, opened 2022-02-08
- Fix quotation for default repo setup

rjschwei:


jgleissner:
Thanks. One comment, the change log generator expects the commit message to be formatted like this:
```
Subject
<NEWLINE>
Detailed changes.
```
It doesn't make sense to start the subject with a hyphen, since it's not supposed to be a list. The detailed changes section can be a list. The log generator actually removes leading '- ' from the subject so it doesn't look odd in the change log. I should add some documentation about the expected format.

jgleissner (approved):

Issue #106, opened 2022-02-08
Overlay tree is not correct for GCE HPC PAYG image


rjschwei:
Should be the same as for the BYOS image. The BYOS image is correct. In the PAYG version limits.d settings are missing for example.

rjschwei:
That said in the BYOS image it looks like `etc/motd.d/30-mgmt` appears to be missing

jgleissner:
`/etc/motd.d/30-mgmt` is only in on-demand images. I think I initially only added SLE-HPC BYOS for GCE as we didn't have an on-demand GCE build in the SUSE namespace and never added the appropriate overlay config for on-demand. I'll fix it.

jgleissner:
Fix was merged.

PR #107, opened 2022-02-09
Include GCE specific config for SLE-HPC 15 in PAYG

jgleissner:
Include GCE specific settings for SLE-HPC 15 also in PAYG descriptions not just BYOS.

Fixes https://github.com/SUSE-Enceladus/keg-recipes/issues/106.


images/cross-cloud/sle-hpc/byos/profiles.yaml
```diff
@@ -19,4 +19,5 @@ profiles:
     description: GCE configuration
     include:
       - csp/gce/addon/gce-tools
+      - csp/gce/addon/hpc-tweaks
```
rjschwei, line 22: hpc-tweaks??? really, how about `hpc-settings` or `hpc-configs`

jgleissner: Well, it tweaks the GCE module for use with the HPC module. I've used the term before for CHOST on Azure, although admittedly, the HPC add-on for GCE is a bit more substantial. I can rename both.

rjschwei (approved):

PR #108, opened 2022-02-09
Include SUSE Manager version in image names

jgleissner:


rjschwei (changes requested):

rjschwei:
And for the image name plaese move the version after `Server` and `Proxy` respectively. Thanks

jgleissner:
> And for the image name plaese move the version after `Server` and `Proxy` respectively. Thanks

Are you sure? This would diverge from the scheme we use for the package names and the cloud image names.

rjschwei:
> > And for the image name plaese move the version after `Server` and `Proxy` respectively. Thanks
> 
> Are you sure? This would diverge from the scheme we use for the package names and the cloud image names.

I guess I screwed that up way back when. `suse-manager-4-1-server-byos-v20220126` does certainly not match the `Product-Version-Flavor` convention. It is probably easier for mashhouse to stick with the busted naming for the internal image as well.

rjschwei (approved):

PR #110, opened 2022-02-15
fix: set schemaversion to 6.2 for Azure LI/VLI 15 SP2

jgleissner:
Released versions use 6.2.

### DIFF
--- a/images/cross-cloud/sles-sap/azure-li/15-sp2/image.yaml
+++ b/images/cross-cloud/sles-sap/azure-li/15-sp2/image.yaml
@@ -1,7 +1,7 @@
 include-paths:
   - sle15/sp1
   - sle15/sp2
-schemaversion: 7.2
+schemaversion: 6.2
 image:
   name: SLES15-SP2-SAP-Azure-LI-BYOS
   specification: "SUSE Linux Enterprise Server for SAP Applications 15 SP2 Azure LI BYOS guest image"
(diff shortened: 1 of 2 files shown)


rjschwei (dismissed):

rjschwei:
@jgleissner but keg is setup to produce data in a specific schema version, I don't quite understand how this changes anything.

jgleissner:
> @jgleissner but keg is setup to produce data in a specific schema version, I don't quite understand how this changes anything.

Keg doesn't really know about schema versions. The structure of the kiwi file is implemented by the template in keg-recipes. The bootloader setup in `data/csp/azure-baremetal` is pre-7.2 style for <=SP2 (analog to the released version) and hence fails with schema version 7.2, which keg really only passes on to the template. The template expects the input data to match the schema version spec.

It's perhaps not as nice as it could be. We could have a template per (supported) schema version that produces the tags how kiwi expects it, and simply just have one standard way to specify bootloader config, which the template translates accordingly. This would also apply to other tags. I'm not sure whether it would be worth the effort.

Another approach might be to use kiwi as the template engine instead of jinja2. I haven't really considered that and I don't know whether it's feasible. AFAIK Marcus has never suggested it so perhaps it's not.

rjschwei:
Thanks for reminding me of this, once again.

rjschwei (approved):

PR #113, opened 2022-03-03
add image definitions for SLES SAPCAL 15 SP3 and SP4

jgleissner:
This adds image definitions for SLES SAPCAL 15 SP3 and SP4. I've created an SP3 test build and compared the images to the released versions. There are a couple of differences between them. Most are changes that have been discussed and agreed on in conjunction with other images, like fixing small inconsistencies (like adding `azure-cli`), getting rid of obsolete stuff (like `/etc/securetty` entry). There are a couple of differences that may need discussion though:

- zypper-lifecycle-plugin (all flavors)

This is included in every image (except CHOST) in `keg-recipes`. It's not in the released SLES SAPCAL 15 SP3 version, but I think it should be.

- rpm-ndb (all flavors)

The description for the released image does not include `rpm-ndb` directly and lacks the related OBS comment. It does not make a difference in the binary image though because the OBS project is set up to switch `rpm` out for `rpm-ndb`. I assume this was just an oversight and the explicit `rpm-ndb` setting should be included in SAPCAL as well.

- /etc/chrony.d/ec2.conf (EC2)

This is missing in the released image but I don't see a reason why it should be excluded.

### DIFF
--- a/data/products/sapcal/sle15/packages.yaml
+++ b/data/products/sapcal/sle15/packages.yaml
@@ -41,7 +41,8 @@ packages:
       - rpm-build
       - samba
       - samba-client
-      - samba-client-32bit
+      - name: samba-client-32bit
+        arch: x86_64
       - sapconf
       - sysstat
       - tack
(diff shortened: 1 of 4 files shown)


rjschwei (approved):

aosthof (approved):
LGTM - at least to my (limited) knowledge

PR #115, opened 2022-03-04
Add image hardening

KeithMnemonic:
This PR adds image definitions to provide "hardened" images for SLES 15 SP3 and SLES-15-SP3-SAP.  These also apply to SLES 15 SP4.   The hardened images contain packages required for the remediation to be run.  

In addition, both the script that performs the remediation and the corresponding invocation of the script have been added.

A new config script section `apply-pcs-hardening.sh` and the corresponding overlay have been added.

### DIFF
--- a/data/scripts/cloud-remediation.sh
+++ b/data/scripts/cloud-remediation.sh
@@ -1,0 +1,3 @@
+# Run the image hardening script and then remove the package after completion
+/usr/bin/bash -e /usr/bin/cloud-remediation.sh
+zypper --non-interactive rm cloud-remediation
--- a/data/base/hardened/sle15/sp3/config.yaml
+++ b/data/base/hardened/sle15/sp3/config.yaml
@@ -1,0 +1,4 @@
+config:
+  scripts:
+    hardened-config:
+      - cloud-remediation
--- a/data/base/hardened/sle15/sp3/packages.yaml
+++ b/data/base/hardened/sle15/sp3/packages.yaml
@@ -1,0 +1,25 @@
+packages:
+  image:
+    hardened:
+      - aide
+      - audit-audispd-plugins
+      - cloud-remediation
+      - firewalld-lang
+      - libfreebl3-hmac
+      - libopenscap25
+      - libpcsclite1
+      - librdkafka1
+      - libsoftokn3-hmac
+      - libxmlsec1-1
+      - libxmlsec1-openssl1
+      - opensc
+      - openscap
+      - openscap-utils
+      - pam_apparmor
+      - pam_pkcs11
+      - pcsc-ccid
+      - pcsc-lite
+      - pcsc-tools
+      - perl-pcsc
+      - scap-security-guide
+      - pcsc-lite
--- a/images/cross-cloud/sles-sap/byos/hardened/15-sp3/image.yaml
+++ b/images/cross-cloud/sles-sap/byos/hardened/15-sp3/image.yaml
@@ -1,0 +1,9 @@
+include-paths:
+  - sle15/sp1
+  - sle15/sp2
+  - sle15/sp3
+schemaversion: 7.2
+image:
+  name: SLES15-SP3-SAP-BYOS-Hardened
+  specification: "SUSE Linux Enterprise Server for SAP Applications 15 SP3 BYOS guest image"
+  version: 0.9.1
--- a/images/cross-cloud/sles-sap/byos/hardened/profiles.yaml
+++ b/images/cross-cloud/sles-sap/byos/hardened/profiles.yaml
@@ -1,0 +1,7 @@
+profiles:
+  common:
+    include:
+      - base/common
+      - base/hardened
+      - base/sle
+      - products/sap/byos
--- a/images/cross-cloud/sles-sap/ondemand/hardened/15-sp3/image.yaml
+++ b/images/cross-cloud/sles-sap/ondemand/hardened/15-sp3/image.yaml
@@ -1,0 +1,9 @@
+include-paths:
+  - sle15/sp1
+  - sle15/sp2
+  - sle15/sp3
+schemaversion: 7.2
+image:
+  name: SLES15-SP3-SAP-Hardened
+  specification: "SUSE Linux Enterprise Server for SAP Applications 15 SP3 guest image"
+  version: 0.9.1
--- a/images/cross-cloud/sles-sap/ondemand/hardened/profiles.yaml
+++ b/images/cross-cloud/sles-sap/ondemand/hardened/profiles.yaml
@@ -1,0 +1,8 @@
+profiles:
+  common:
+    include:
+      - base/common
+      - base/hardened
+      - base/sle
+      - base/sle-modules
+      - products/sap/ondemand
--- a/images/cross-cloud/sles/byos/hardened/15-sp3/image.yaml
+++ b/images/cross-cloud/sles/byos/hardened/15-sp3/image.yaml
@@ -1,0 +1,9 @@
+include-paths:
+  - sle15/sp1
+  - sle15/sp2
+  - sle15/sp3
+schemaversion: 7.2
+image:
+  name: SLES15-SP3-BYOS-Hardened
+  specification: "SUSE Linux Enterprise Server 15 SP3 BYOS guest image"
+  version: 0.9.1
--- a/images/cross-cloud/sles/byos/hardened/profiles.yaml
+++ b/images/cross-cloud/sles/byos/hardened/profiles.yaml
@@ -1,0 +1,7 @@
+profiles:
+  common:
+    include:
+      - base/common
+      - base/hardened
+      - base/sle
+      - products/sles/byos
--- a/images/cross-cloud/sles/ondemand/hardened/15-sp3/image.yaml
+++ b/images/cross-cloud/sles/ondemand/hardened/15-sp3/image.yaml
@@ -1,0 +1,9 @@
+include-paths:
+  - sle15/sp1
+  - sle15/sp2
+  - sle15/sp3
+schemaversion: 7.2
+image:
+  name: SLES15-SP3-Hardened
+  specification: "SUSE Linux Enterprise Server 15 SP3 guest image"
+  version: 0.9.1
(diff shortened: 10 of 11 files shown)


images/cross-cloud/sles/byos/profiles.yaml
```diff
@@ -3,7 +3,6 @@ profiles:
     include:
       - base/common
       - base/sle
-      - products/sles/byos
```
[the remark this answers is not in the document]
KeithMnemonic: Thanks for catching that, I did not intend to remove it. I push as new commit fixes this and does not remove it

KeithMnemonic:
> Another thing, `/opt/pct/pcs-hardening.sh` is an astonishing 31909 lines of generated shell script. I'm not sure whether keeping it in `keg-recipes` would be ideal in terms of maintainability. I don't know what source data it operates on, but wouldn't it need to be re-generated whenever the source data changes? And could the script differ between SPs?

The script could differ if a new SP needed additional remediation scripts than a prior SP. At this point we have only been tasked with doing 15-SP3.  Is there a way in keg to "inherit" and overly an then append on top of it?    I.e have a bae script for 15-SP3 and then additional scripts  for 15-SP4, 15-SP5 for example .The question about source data can be answered by Brett. I am not sure if he has access here.

jgleissner:
> The script could differ if a new SP needed additional remediation scripts than a prior SP. At this point we have only been tasked with doing 15-SP3. Is there a way in keg to "inherit" and overly an then append on top of it? I.e have a bae script for 15-SP3 and then additional scripts for 15-SP4, 15-SP5 for example .The question about source data can be answered by Brett . I am not sure if he has access here.

Overlay files and scriptlets don't have the inheritance and versioning mechanism of the yaml data. We decided to keep it simple, and the yaml config that references those files has those mechanisms so it would be possible to simply provide different versions as different overlay files for different SPs. The overlay files are intended for small bits of mostly static data, it's not really designed for a file that would need to be regenerated regularly. This could be done as a package build or something. If we only need to do it once in a blue moon it may not be worth it, although I think the less manual stuff we need to do the better.

rjschwei:
> @rjschwei I thought during one of our recent hardening meetings, you had indicated I should implement this as an additional profile for the image. i.e Azure and Azure-Hardened. IIUC, this is not what @jgleissner is suggesting when he mentions. "Since this results in duplication anyway, I don't really see an advantage of having the hardened images in the same image definitions as the standard ones." I think this method suggested from Joe would mean a new IBS project? If I am understanding Joes suggestion correctly, can we come up with a definitive answer on how this should be implemented. Do we do profiles or a seperate image definition

@KeithMnemonic it would not mean a new project in IBS, but following @jgleissner path means a new "package", i.e. a net new image description. We'd have `SLES15-SP3-SAP-Hardened`, `SLES15-SP3-SAP-BYOS-Hardened` and `SLES15-SP3-Hardened` each of these have 3 profiles `Azure, EC2, GCE`

This does cut down on the duplication as in keg/keg-recipes we do not have combination logic. In a hand crafted config file we could for example have

``` <packages type="image" profiles="EC2,EC2_Hardened">
     <package name="cloud-init"/>
     ...
    </packages>
    <packages type="image" profiles="Azure_Hardened,EC2_Hardened,GCE_Hardened">
    <package name="foo"/>
    .....
    </packages>
```

But in keg we do not have such combination options and as such all of this would be duplicated 3 times over. But if we create dedicated image descriptions as @jgleissner suggests we minimize the duplication. Given that we can still inherit from the common package definitions and the hardened packages become an addon the maintenance is still reasonably simple. Thus we should follow @jgleissner idea and create `SLES15-SP3-SAP-Hardened`, `SLES15-SP3-SAP-BYOS-Hardened` and `SLES15-SP3-Hardened`

rjschwei:
@KeithMnemonic @jgleissner I think the remediation script should live outside of keg-recipes and it should come in an rpm package that we include in the Public Cloud Module and publish to our customers in the repo.

We can set up a packaging trick such that in keg-recipes we have `<package name="cloud-remediation"/>` and on the backend we pick up an SP specific version of the script. This type of version switching is implemented for `libyui` if an example is.needed.

KeithMnemonic:
> > @rjschwei I thought during one of our recent hardening meetings, you had indicated I should implement this as an additional profile for the image. i.e Azure and Azure-Hardened. IIUC, this is not what @jgleissner is suggesting when he mentions. "Since this results in duplication anyway, I don't really see an advantage of having the hardened images in the same image definitions as the standard ones." I think this method suggested from Joe would mean a new IBS project? If I am understanding Joes suggestion correctly, can we come up with a definitive answer on how this should be implemented. Do we do profiles or a seperate image definition
> 
> @KeithMnemonic it would not mean a new project in IBS, but following @jgleissner path means a new "package", i.e. a net new image description. We'd have `SLES15-SP3-SAP-Hardened`, `SLES15-SP3-SAP-BYOS-Hardened` and `SLES15-SP3-Hardened` each of these have 3 profiles `Azure, EC2, GCE`
> 
> This does cut down on the duplication as in keg/keg-recipes we do not have combination logic. In a hand crafted config file we could for example have
> 
> ```
>      <package name="cloud-init"/>
>      ...
>     </packages>
>     <packages type="image" profiles="Azure_Hardened,EC2_Hardened,GCE_Hardened">
>     <package name="foo"/>
>     .....
>     </packages>
> ```
> 
> But in keg we do not have such combination options and as such all of this would be duplicated 3 times over. But if we create dedicated image descriptions as @jgleissner suggests we minimize the duplication. Given that we can still inherit from the common package definitions and the hardened packages become an addon the maintenance is still reasonably simple. Thus we should follow @jgleissner idea and create `SLES15-SP3-SAP-Hardened`, `SLES15-SP3-SAP-BYOS-Hardened` and `SLES15-SP3-Hardened`

@rjschwei  Thanks, I have this working now with Joe's help.  Are we omitting `SLES15-SP3-BYOS-Hardened` or was that by accident you did not mention it?

rjschwei:
@KeithMnemonic it should have been `SLES15-SP3-BYOS-Hardened` and not `SLES15-SP3-Hardened`. Due to the qualification of such images there is a life-cycle problem for SLES PAYG and as such that is currently not being considred for release. As far as keg is concerned, yes we need to be able to create all 4 image descriptions SLES4SAP PAYG & BYOS and SLES PAYG & BYOS

KeithMnemonic:
@jgleissner  Please review again and see if this looks good now. @rjschwei  feel free to comment as well.   If all is good, I guess we hold it until we have a working cloud-remediation package in `Update:` ?

jgleissner (changes requested):

jgleissner (approved):
LGTM

data/scripts/cloud-remediation.sh
```diff
@@ -0,0 +1,3 @@
+# Run the image hardening script and then remove the package after completion
+/opt/pct/cloud-remediation.sh
```
rjschwei, line 2: This makes the assumption that the script is set to executable, which is fair to a certain extend. That said it might be safer to specify the interpreter, `/usr/bin/bash` if this is a bash script, or `/usr/bin/sh` or whatever.

The other question we need to answer if we want the image build to fail if application of remediation fails. Then we'd have to check the exit condition and exit.

rjschwei (approved):

data/scripts/cloud-remediation.sh
KeithMnemonic: I can add the interpreter path to the call.  Do you want to make the call to the image failing if remediation fails or do we discuss that in a larger group setting?

rjschwei: I think it should fail, we are trying to produce a hardened image after all. @jgleissner thoughts?

KeithMnemonic: see if this is works for what you are expecting

data/scripts/cloud-remediation.sh
```diff
@@ -0,0 +1,6 @@
+# Run the image hardening script and then remove the package after completion
+/usr/bin/bash -e /opt/pct/cloud-remediation.sh
+if [[ $? -ne 0 ]] ; then
+    exit 1
+fi
```
jgleissner, line 5: Checking the exit status is not necessary. `config.sh` sets the abort-on-error flag. I'm not sure about running `cloud-remediation.sh` itself with `-e`. This could have side-effects. I'd be reluctant to force it on external scripts.

KeithMnemonic: I will adjust per your comments push it up shortly

KeithMnemonic: In testing I think we need the -e. The remediation script which is generated from here 

https://github.com/brett060102/content/tree/pcs-hardening/products/sle15/profiles

does not use exit codes in case a remediation is not applied as this could mean there was nothing to do.  I tested this with and without the -e and in only aborts correctly when the -e is present

jgleissner: A had a look at the script that's in your IBS project and it uses `false` statements to indicate error situations without forcing abort, so it seems it's actually designed to be run with `-e`. I just mentioned possible side-effects because if a script isn't designed for that, it may break, but in this case I think we do want to run it with `-e`.

As a side note, wouldn't it be better to generate the script in the build instead of simply packaging the script file?

KeithMnemonic: Joe, the idea is the cloud-remediation is only needed until this https://github.com/brett060102/content/tree/pcs-hardening get merged upstream.  We have no ETA at to that time frame. Once that is done however, the profiles listed here https://github.com/ComplianceAsCode/content are able to be remediated using the oscap tools.  When that occurs, we simply update the scriptlet in keg recipes. I do not think it would be optimal to have the script generated as part of he cloud-remediation  package build, but that conversation can take place when iI do the 'sr' for cloud-remediation.  Any reason not to merge this change here now.?

jgleissner: Ah ok, I didn't realize this was just a temporary thing. In that case we might as well keep it in `keg-recipes` for the time being. If that package just contains the generated script, there isn't really a benefit IMO. Anyhow, PR is fine for merge from my POV.

PR #133, opened 2022-07-05
Convert all definitions to 

jgleissner:


jgleissner:
Sorry for the lack of an initial comment, that was an accident. This PR converts all data to the new `minikeg` format. All image definitions should generate and verify, but I expect plenty of small things that need fixing. I'll set up a full array of test builds.

This rearranges the data modules in a way that I hope feels less cluttered and more consistent. I think the approach works well for most image definitions except SLE 12 SP5. The data layout isn't really geared towards generating multiple products from one description but one product for multiple platforms, as later image definitions use it. This leads to quite a bit of duplication, especially in the generated data. With old keg we used nested profiles to mitigate it, but since keg profiles are gone, that's not possible anymore. I see the following options:

1. Rearrange SLE12 data

SLE 12 and SLE 15 data are already completely separate so we could choose a different layout that accommodates the way SLE 12 SP5 image descriptions work better. Downside is that having two different data layouts for SLE 12 and SLE 15 may make working with the data a bit confusing.

2. Implement some sort of selective include

When generating sub profiles, old keg would check against the base profile for existing keys and only include new keys, which allowed to e.g. include `csp/azure` in the base profile and `csp/azure/foo` in the sub profile, and not end up with keys from `csp/azure` in the sub profile. If we add a way to allow the user to exclude certain top level dirs, the nested profile mechanism could somewhat be replicated.

3. Accept that SLE12 SP5 descriptions have unnecessary duplication

Well, it is an option. ;)

rjschwei:
Option 4:

Cast SLE 12 SP5 images into the image description per product format.

rjschwei:
I think this looks reasonably close. What I am missing is the incorporation of the architecture specific profilei.e. we need to have

```
<preferences profiles="Azure-Standard" arch="aarch64">
```
generated. As such the setup for aarch64 in Azure is missing. This can be handled in a follow up PR but I'd like to see that included before we move toward pushing this new setup to the main branch.

rjschwei (approved):

jgleissner:
I meant to include the arch specific profiles, but forgot, sorry. It can be implemented fairly easily.

I'll open a separate PR for this, and also for whatever we decide to do with the 12 SP5 definitions.

PR #136, opened 2022-07-26
More fixes for "minikeg"

jgleissner:
Collection of fixes for `minikeg` definitions.

### DIFF
--- a/data/csp/azure/sle12/packages.yaml
+++ b/data/csp/azure/sle12/packages.yaml
@@ -23,7 +23,6 @@ packages:
     package:
       - azure-cli
       - lsscsi
-      - python3-azure-sdk
       - python3-azuremetadata
   _namespace_azure_registration:
     package:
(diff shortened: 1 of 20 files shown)


data/csp/azure/sle12/packages.yaml
```diff
@@ -23,7 +23,6 @@ packages:
     package:
       - azure-cli
       - lsscsi
-      - python3-azure-sdk
```
rjschwei, line 26: Should be `python-azure-sdk`

rjschwei (approved):

data/csp/azure/sle12/packages.yaml
jgleissner: I don't see `python-azure-sdk` included at all in SLE 12 so that's why I simply removed `python3-azure-sdk`, which I added accidentally in the first place.

rjschwei: Only added it to the 15 images but we should include it in the 12 based images as well since we are shipping the packages and we've sorted out the various issues we had early on.

jgleissner: Ok, easy enough to add.

PR #138, opened 2022-08-04
Add hardened for 15 sp4

KeithMnemonic:
This PR extends the image hardening that was added for 15-sp3 to 15-sp4.


jgleissner:
This shouldn't be a PR into `released` but `develop`. But since we're changing basically everything (see branch [minikeg-staging](https://github.com/SUSE-Enceladus/keg-recipes/tree/minikeg-staging) there isn't much point in merging this anyway, unless perhaps it's urgently needed. I'll handle the `minikeg` conversion of the hardened image definitions.

KeithMnemonic:
> This shouldn't be a PR into `released` but `develop`. But since we're changing basically everything (see branch [minikeg-staging](https://github.com/SUSE-Enceladus/keg-recipes/tree/minikeg-staging) there isn't much point in merging this anyway, unless perhaps it's urgently needed. I'll handle the `minikeg` conversion of the hardened image definitions.

@jgleissner Joe, thanks for catching this.  Thank you for offering to do this is mini-keg. Since we are not releasing for 15-SP3 and only for 15-SP4+ maybie it makes sense to remove the 15-SP3 entries?

rjschwei:
With everything moving to mini-keg I will close this.

PR #139, opened 2022-08-05
Even more fixes for minikeg

jgleissner:
This PR contains fixes to bring the minikeg image definitions in line with the released image descriptions. There will almost certainly be more but since they have accumulated quite a bit already I figured it'd be better to do this in several PRs.

With this PR, SLE 15 SP4 descriptions should match the released ones, with a couple of exceptions, namely the removal of `klogd` where previously included and the addition of the work-around for bsc#1197616.


images/cross-cloud/sle-micro/byos/content.yaml
```diff
@@ -41,23 +41,23 @@ image:
...
+        - csp/ec2/settings/micro-os
       archive:
         - _attributes:
             name: ec2.tar.gz
     - _attributes:
         type: image
         profiles: [GCE]
       _include:
-        - csp/gce
+        - csp/gce/settings/micro-os
```
rjschwei, line 60: We are not supposed to use the `micro-os` name

rjschwei (approved):

rjschwei:
Merging, we'll still want to do a rename for `micro-os` in a follow up PR

PR #140, opened 2022-08-15
SLE Micro fixes for minikeg

jgleissner:
This brings the SLE image definition mostly in line with the existing ones, with a few of exceptions:

- `hyper-v` added to Azure aarch64 (not included in released)
- `klogd` removed (not supported)
- zypper-migration-plugin (should this not be included?)
- cloud-netconfig enabled (package included in released image but not enabled)
- kernel multiversion disabled in `zypp.conf`


data/products/sle-micro/config.yaml
```diff
@@ -1,17 +1,22 @@
 config:
   scripts:
-    sle-micro:
+    sle-micro-scripts:
       - chost-strip-image
       - microos-add-variant
```
rjschwei, line 5: Another reference to `microos`

data/products/sle-micro/config.yaml
```diff
@@ -1,17 +1,22 @@
...
       - snapper-microos-config
```
rjschwei, line 6: ditto

data/products/sle-micro/config.yaml
```diff
@@ -1,17 +1,22 @@
...
       - afterburn-checkin
       - afterburn-firstboot-checkin
       - afterburn-sshkeys@
       - afterburn-sshkeys.target
       - boot.device-mapper
       - docker
+  sysconfig:
+    sle-micro-sysconig:
+      - file: /etc/sysconfig/network/config
+        name: WICKED_DEBUG
```
rjschwei, line 21: SLe Micro 5.3 drops wicked and moves to NetworkManager, this will need to be version specific.

rjschwei (approved):
The comments should/can be addressed in a follow up PR

PR #146, opened 2022-08-25
Restructure SLES 12 SP5 Azure image definition

jgleissner:
This restructures some of the SLE12 data and changes the SLES 12 SP5 Azure image definition to almost completely eradicate data duplication in the produced image description.

Since the data modules are designed to work in one image definition per product, they don't work too nicely with the all-products-in-one image definition for SLES 12 SP5. I initially rewrote the whole products subtree for SLE12, but wasn't happy with it and it turned out to be not really necessary. Instead I ended up only using the base product modules, so they can be shared with all relevant profiles, and added the few additional bits (e.g. BYOS or PAYG specific ones) in the image definition directly, without even adding a separate module for them (which you can actually do in minikeg contrary to original keg).

I think this is a decent solution as it allows to replicate the existing SLES 12 SP5 descriptions fairly faithfully without disturbing the data module layout that is geared towards the newer description profile.

If this is acceptable, I'd convert the EC2 and GCE definitions as well.


rjschwei (approved):

rjschwei:
I think this works and also shows the higher degree of flexibility of the minikeg approach.

PR #147, opened 2022-08-26
Include hostinfo and address iscsi config

rjschwei:
- Include the hostinfo package in our images by default per AWS request
  (bsc#1199872). This also supports the expansion of information in motd
- Remove the iscsi initiator config that gets generated during package install
  to avoid the same initiator name on all instances. While the change was
  introduced in 15 SP4 there is no harm in having the remove in all image
  builds. Therefore this is included in the default setup for SLE 15 and
  SLE 12. (bsc#1202540)


jgleissner (approved):
LGTM

PR #148, opened 2022-08-29
Ignore 'rpm' (use rpm-ndb) also for single-cloud descriptions

aosthof:
The comment `<!-- OBS-IgnorePackage: rpm -->` is not used for single-cloud image description and currently results in unresolvables, see https://build.suse.de/project/monitor/SUSE:SLE-15-SP5:GA:Staging:B?arch_x86_64=1&defaults=0&repo_images=1&unresolvable=1 for instance.

### DIFF
--- a/images/single-cloud/obs-comments.yaml
+++ b/images/single-cloud/obs-comments.yaml
@@ -1,0 +1,2 @@
+image-config-comments:
+  obs-ignore-rpm: "OBS-IgnorePackage: rpm"


rjschwei (approved):

jgleissner:
Thanks for spotting that. Background is that `cross-cloud` and `single-cloud` evolved from something like current scheme and old scheme. With moving some of the newer definitions to `single-cloud` where they actually belong, the `IgnorePackage` directive wasn't applied anymore. However, this PR enables it for all older image definitions which is obviously not intended. I'll fix it.

PR #151, opened 2022-09-14
Minikeg image definition updates

jgleissner:
This PR contains mostly the restructure of the SLE12-SP5 image definition to avoid most of the duplication in the generated descriptions. There are also a few additional fixes.

I ran test builds and compared the images and they are mostly in line of the released ones, with a couple of exceptions:

All images had `klogd` dropped as it wasn't released for SLE 12 and only in the dev tools module in SLE 15.

`hostinfo` was added to all images (including Azure LI/VLI).

SLES12-SP4-SAP-Azure now has `cloud-regionsrv-client-addon-azure` and the associated service and timer enabled. It's in the SLES 12 SP5 image so I figured it's supposed to be in SP4 as well.

SLES12-SP5-EC2-ECS-On-Demand has `libnl1` which is not in the released version. The package is included in all other SLES 12 images, and hence is in `base/common`. I can shift it around if it's not ok to include it (it's small and does not have dependencies).

Azure LI/VLI has `sle-module-public-cloud-release` added. We added that to all images, but I'm not sure whether it's intended for those.


jgleissner:
@rjschwei FYI, that's the last PR I have for minikeg. I think we could go ahead and make the switch once this is merged.

data/csp/azure/sle12/packages.yaml
```diff
@@ -1,24 +1,20 @@
...
-    package:
-      - _attributes:
-         name: kernel-default
-         arch: x86_64
   _namespace_azure_cloud_netconfig:
     package:
       - cloud-netconfig-azure
+  _namespace_azure_kernel:
+    package:
+      - kernel-default
```
rjschwei, line 17: One would think that `_namespace_azure_kernel` should have `kernel-azure` but this was already weird previously.

rjschwei (approved):

rjschwei:
Let's go for it and merge mini-keg

data/csp/azure/sle12/packages.yaml
jgleissner: Well it's not the namespace for `kernel-azure` but the namespace for whatever kernel is used in Azure, which is `kernel-default` by default and redefined in settings modules that use `kernel-azure`. But I can see how it may be confusing. ;)

PR #160, opened 2022-10-03
Disable sshd service for Rancher Setup

bear454:


rjschwei (approved):

data/products/rancher-setup/sle15/config.yaml
```diff
@@ -4,3 +4,6 @@ config:
       - suse-rancher-setup
       - nginx
       - set-http-basic-credentials
+    common-services:
+      - name: sshd
+        enable: false
```
jgleissner, line 9: Overwriting `common-services` here isn't ideal because it's not defined in the `products` tree but is included from `base/common`. Overwriting namespaces that aren't directly inherited only works when those modules are included at the same time and only in the right order. That could be a bit confusing as it means you'd get different results depending on how you include it and order becomes significant too, so I generally don't use it.

It also removes other service settings in `base/common/sle15/config.yaml`, which are admittedly all without actual effect and we could even consider dropping them. We could also consider using a systemd preset file instead of enabling services in `config.sh`, which would make it a bit more transparent to the user what the defaults are I think.

Anyhow, I'd suggest moving sshd from `base/common/sle15/config.yaml` to its own namespace in `products/sle15/defaults.yaml` and overwrite that one instead.

jgleissner (changes requested):

data/products/rancher-setup/sle15/config.yaml
bear454: Like a93f345 ?

PR #163, opened 2022-10-13
Add OCI image info

smarlowucf:


data/csp/oci/sle15/packages.yaml
```diff
@@ -0,0 +1,13 @@
+packages:
+  image:
+    OCI:
+      - firewalld
+      - libselinux1
+      - open-iscsi
+      - selinux-tools
+      - cloud-init
+drivers:
```
smarlowucf, line 9: Not sure if this is how to handle the drivers?

smarlowucf:
The config.sh has a line "# Need to allow modules from Enterprise Build Service". This same line exists for the Azure image but I don't see it handled in Keg so I left it out for now in case it's no longer needed?

smarlowucf:
Didn't do any testing yet.

jgleissner:
Thanks for diving in. It appears you've used the old (pre-"minikeg") format though. Maybe your keg-recipes repo was out-of-date?

jgleissner:
> The config.sh has a line "# Need to allow modules from Enterprise Build Service". This same line exists for the Azure image but I don't see it handled in Keg so I left it out for now in case it's no longer needed?
> 
> Full snippet:
> 
> ```
> # Need to allow modules from Enterprise Build Service
>     if [ -f /etc/modprobe.d/unsupported-modules ];then
>         sed -i -r -e 's/^(allow_unsupported_modules[[:space:]]*).*/\10/' \
>             /etc/modprobe.d/unsupported-modules
>     fi
> ```

There is no file `/etc/modprobe.d/unsupported-modules` (or at least hasn't been in a while) so this was essentially a no-op. Since this never caused any complaints we simply dropped this.

data/csp/oci/sle15/packages.yaml
jgleissner: Format would be:
```
drivers:
  _namespace_foo:
    file:
      - _attributes:
          name: ixgbe
      ...
```
Which then can be included in a `drivers` section in the image definition.

data/overlayfiles/cloud-cfg-oci/etc/cloud/cloud.cfg
```diff
@@ -27,6 +27,7 @@ cloud_init_modules:
 # The modules that run in the 'config' stage
 cloud_config_modules:
  - mounts
+ - ssh-import-id
```
jgleissner, line 30: `ssh-import-id` was deliberately removed as this was never supported.

smarlowucf:
> Thanks for diving in. It appears you've used the old (pre-"minikeg") format though. Maybe your keg-recipes repo was out-of-date?

Yeah, I thought I was looking at devel branch but it's on released :facepalm:

smarlowucf:
@jgleissner thanks for review. I updated everything to match the new format (I think) and incorporated comments. Squashed commits for cleaner history.

smarlowucf:
> Which then can be included in a drivers section in the image definition.

By definition you mean the config.yaml or still in packages.yaml?

jgleissner (changes requested):

jgleissner:
> > Which then can be included in a drivers section in the image definition.
> 
> By definition you mean the config.yaml or still in packages.yaml?

The image definition in `images/`. All images are defined there and pull in modules from `data/` via `_include` statements. You didn't add image definitions for OCI images with the PR, which is actually fine for the moment, as we need to decide how to do that first anyway. We had OCI profiles in SP2 and SP3 image descriptions, but I don't think we want to add them as profiles to the existing keg recipes as we don't actually plan on releasing new OCI images. We could add a single build definitions that people could use to create image descriptions themselves, and at the same time keep OCI data out of ours. @rjschwei Any opinion on that?

smarlowucf:
Ah, okay I think I follow now on the definitions.

rjschwei:
> > > Which then can be included in a drivers section in the image definition.
> > 
> > 
> > By definition you mean the config.yaml or still in packages.yaml?
> 
> The image definition in `images/`. All images are defined there and pull in modules from `data/` via `_include` statements. You didn't add image definitions for OCI images with the PR, which is actually fine for the moment, as we need to decide how to do that first anyway. We had OCI profiles in SP2 and SP3 image descriptions, but I don't think we want to add them as profiles to the existing keg recipes as we don't actually plan on releasing new OCI images. We could add a single build definitions that people could use to create image descriptions themselves, and at the same time keep OCI data out of ours. @rjschwei Any opinion on that?

I am in favor of setting OCI up as single build. While we plan to eventually use generated definitions for images that are still maintained in SP3, SP2, and SP1 we will no longer publish images in OCI, as such dropping OCI as a profile makes sense. Additionally since we will be directing the build OCI images specifically here it makes sense to keep the profile stuff out of those user's hair as well. People that want to build their own images and need multicloud can easily add other profiles to their own branch of keg-recipes.

data/scripts/iscid-setup.sh
```diff
@@ -0,0 +1,3 @@
+sed -i 's/node.session.timeo.replacement_timeout/# node.session.timeo.replacement_timeout' /etc/iscsi/iscsid.conf
+sed -i 's/node.conn[0].timeo.noop_out_interval/# node.conn[0].timeo.noop_out_interval' /etc/iscsi/iscsid.conf
+sed -i 's/node.conn[0].timeo.noop_out_timeout/# node.conn[0].timeo.noop_out_timeout' /etc/iscsi/iscsid.conf
```
jgleissner, line 3: /iscid-setup.sh/iscsid-setup.sh/

data/csp/oci/sle15/overlayfiles.yaml
```diff
@@ -0,0 +1,7 @@
+archive:
+  _namespace_oci_base:
+    _include_overlays:
+      - module-iscsi
+      - cloud-cfg-oci
+      - chrony-oci-ntp
```
jgleissner, line 6: `chrony-oci-ntp` doesn't exist in `overlayfiles`. I think it did at some point but I removed it as it wasn't used.

jgleissner (changes requested):

rjschwei:
> The last commit simply moved the `drivers` section instead of moving it to its own file. We'll still need it, I assume (I'm not actually sure what it does and it doesn't appear to be documented).

`drivers` keeps certain drivers in the initrd that would otherwise get lost. As such it is a way to ensure the image boots on the target deployment system

data/csp/oci/sle15/config.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - chronyd
+      - cloud-init-local
+      - cloud-init
+      - cloud-config
+      - cloud-final
+      - firewalld
+      - iscsid.service
+      - iscsiuio.service
+  drivers:
+    _namepsace_oci:
```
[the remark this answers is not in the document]
smarlowucf: I forgot to add the new file to commit.

smarlowucf:
I guess we maybe want 12 SP5 definitions too in addition to 15 SP3.

data/csp/oci/sle15/drivers.yaml
```diff
@@ -0,0 +1,11 @@
+drivers:
+  _namespace_oci:
+    file:
+      - _attribute
+        name: ixgbe
+    file:
```
jgleissner, line 6: There should be only one line `file:`.

smarlowucf: Okay, so all drivers are attrs of one file?

rjschwei:
> I guess we maybe want 12 SP5 definitions too in addition to 15 SP3.

Yes, thanks

data/csp/oci/sle15/drivers.yaml
jgleissner: `file` is a list and as such produces an XML element for each list element, i.e.:
```
<file name="ixgbe"/>
<file name="ixgbevf"/>
```
Also, and sorry for not noticing earlier, it's `_attributes` and the `:` is missing too. `name` needs to be indented one more step as it's a child of `_attributes`. Alternatively, this can also be a flat list  i.e.:
```
    file:
      - ixgbe
      - ixgbevf
      ...
```
That would require `_map_attribute` to be set in `drivers` somewhere, ideally in https://github.com/SUSE-Enceladus/keg-recipes/blob/develop/data/packages_defattr.yaml, analog to the `packages` version. The `_map_attribute` directive tells the XML generator when encountering a list of strings to produce a dict for each element with a `_attributes` key, and add a child element with the given name and the string as its value. This allows to makes those lists more compact and readable.

jgleissner: Hey @smarlowucf , any update on this? The issue with the driver attributes is the only thing that's left before this can be merged AFAICS.

smarlowucf: Pushed a commit. I think maybe I figured it out. :crossed_fingers:

data/scripts/iscsid-setup.sh
```diff
@@ -0,0 +1,3 @@
+sed -i 's/node.session.timeo.replacement_timeout/# node.session.timeo.replacement_timeout' /etc/iscsi/iscsid.conf
+sed -i 's/node.conn[0].timeo.noop_out_interval/# node.conn[0].timeo.noop_out_interval' /etc/iscsi/iscsid.conf
+sed -i 's/node.conn[0].timeo.noop_out_timeout/# node.conn[0].timeo.noop_out_timeout' /etc/iscsi/iscsid.conf
```
jgleissner, line 3: The sed commands are lacking the final `/`. The dots would technically need to be escaped so they don't match any char, but this won't make a difference in practice. Bonus nit-pick: Those three sed executions could be compressed into one with three `-e` parameters.  ;-)

jgleissner (changes requested):

data/scripts/iscsid-setup.sh
rjschwei: And I can write an entire perl script in one line as well.

jgleissner: Nobody said it needs to be one line. ;-) It's not important, of course. We still need those slashes though.

jgleissner (approved):
Looks good, thanks!

rjschwei (approved):

PR #166, opened 2022-10-18
Add package to resolve file dependency

rjschwei:
SAPHanaSR has a file dependency on /usr/bin/xmllint but file dependencies cannot be resolved in the build service. Add the package to get proper dependency resolution during build.

### DIFF
--- a/data/products/sap/sle12/packages.yaml
+++ b/data/products/sap/sle12/packages.yaml
@@ -20,6 +20,7 @@ packages:
       - libopenssl0_9_8
       - libpng12-0
       - libvirt-client
+      - libxml2-tools
       - libyui-qt6
       - ocfs2-kmp-default
       - patterns-ha-ha_sles
--- a/data/products/sap/sle15/packages.yaml
+++ b/data/products/sap/sle15/packages.yaml
@@ -20,6 +20,7 @@ packages:
       - libgtk-2_0-0
       - libjpeg62
       - libpng12-0
+      - libxml2-tools
       - libyui-qt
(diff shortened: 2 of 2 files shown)


jgleissner:
AFAICS the build service resolves file dependencies just fine and `libxml2-tools` is included in all SAP builds already, so is this really needed?

rjschwei:
> AFAICS the build service resolves file dependencies just fine and `libxml2-tools` is included in all SAP builds already, so is this really needed?

I did not check if `libxml2-tools` is already included based on some other component.

In the develop branch

```
-> find . -type f | xargs grep libxml2-tools
->
```
So I am not sure how """`libxml2-tools` is included in all SAP builds already,""" would be a true statement.

 I would say the build failures in 12 SP4 and SP5 that I fixed by adding this package to the image description is evidence that the build service does not resolve a file dependency such as `Requires:       /usr/bin/xmllint`. 

Based on the symptoms (builds fell over for 12 SP4 & SP5) I would say yes that is needed. I did no check how the HANASr package declares it's dependencies in the SLE 15 branch, maybe that is different.

jgleissner:
libxml2-tools is included in the (SLES 15) SAP builds via dependency, it's not in keg-recipes. I hadn't checked SLES 12, and in those builds `libxml2-tools` is indeed not added automatically. Which seems odd, file based dependencies have been around forever. Both SLES 12 and SLES 15 versions of SAPHanaSR require `/usr/bin/xmllint` directly. Anyhow, this PR will work around it, of course.

PR #171, opened 2022-11-02
Fix variant IDs for hardened definitions

jgleissner:
Use 'sles-hardened' for hardened SLES.
Use 'sles-sap-hardened' for hardened SLES for SAP.

### DIFF
--- a/data/products/sap/hardened/sle15/config.yaml
+++ b/data/products/sap/hardened/sle15/config.yaml
@@ -1,4 +1,9 @@
 config:
+  files:
+    sles4sap-variant:
+      - path: /etc/os-release
+        append: True
+        content: VARIANT_ID="sles-sap-hardened"
   scripts:
     hardened-config:
       - pcs-sap-hardening
--- a/data/base/sle/hardened/sle15/config.yaml
+++ b/data/base/sle/hardened/sle15/config.yaml
@@ -1,6 +1,0 @@
-config:
-  files:
-    hardened_variant_id:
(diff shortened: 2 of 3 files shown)


rjschwei (approved):

data/products/sap/hardened/sle15/config.yaml
```diff
@@ -1,4 +1,9 @@
 config:
+  files:
+    sles4sap-variant:
```
aosthof, line 3: Not a blocker but cosmetic: I'd replace "sles4sap-variant" with "sles-sap-hardened-variant-id" to be consistent with the "sles-hardened-variant-id" name.

aosthof (commented):
Rest looks good to me.

data/products/sap/hardened/sle15/config.yaml
jgleissner: This overwrites the `sles4sap-variant` namespace from `data/products/sap/sle15/config.yaml`. We could null it instead and add a namespace `sles-sap-hardened-varian-id` if you prefer. I did that in other places where the resulting comment in config.kiwi seemed inappropriate otherwise but I didn't think it mattered in this case.

aosthof:
As Joe is on vacation now I'll approve this PR so we can move on with the hardened images. The cosmetic issue can also be fixed at a later point in time if needed/wanted.

aosthof (approved):
LGTM

PR #173, opened 2022-11-16
Add package 'ed' to Azure LI/VLI images

aosthof:
As the requirements of the package 'postfix' have changed with SR#280878 we need to add the package 'ed' to the Azure LI/VLI images to fulfill the dependency. See also [bsc#1205457](https://bugzilla.suse.com/show_bug.cgi?id=1205457).


jgleissner (approved):

rjschwei:
Do we know which package drags postfix into the images? We do not explicitly include postfix. The inclusion should be in the block of the package that brings postfix along for the ride. If it is `fipscheck` we're OK, but we should have a better commit message to indicate the dependency.

aosthof:
I don't know how postfix is dragged into the image. But the error message in bsc#1205457 lists "Solution 1: do not install patterns-sap-hana-15.4-150500.7.1.x86_64" so I assume one of the packages listed in the pattern has a dependency on postfix. The package postfix itself is not part of the pattern though.

aosthof:
I need to rework the PR, therefore closing this one and starting new.

aosthof:
I need to get more familiar with handling PRs :d'oh: reopening.

aosthof:
- ed is required by postfix
- postfix is required by libmilter1_0
- libmilter1_0 is required by clamav
- clamav is required by clamsap
- clamsap is required by patterns-sap-hana

rjschwei (approved):

PR #174, opened 2022-11-17
Move 'ed' to the generic sle15 packages.yaml to avoid namespace clashes

aosthof:
When I added the package 'ed' in data/csp/azure-baremetal/sle15/sp5/packages.yaml I used the same namespace as in data/csp/azure-baremetal/sle15/packages.yaml which overwrote the latter, so packages were missing. After discussion this issue with Joe we agreed on creating the new namespace '_namespace_azure_baremetal_hana_addon_deps' to solve this issue.


jgleissner (approved):
Doesn't look that different from the original PR. ;) Thanks, LGTM.

rjschwei (approved):

PR #175, opened 2022-11-18
Add single build OCI 15 sp3 definition

smarlowucf:


images/single-cloud/oci/sles-sap-byos/content.yaml
```diff
@@ -0,0 +1,30 @@
...
+  packages:
+    - _attributes:
+        type: bootstrap
+      _include: base/bootstrap
+    - _attributes:
+        type: image
+      _include:
+        - base/common
+        - base/sle
+        - base/byos
```
jgleissner, line 15: There's no `base/byos`, should be removed.

jgleissner (changes requested):

images/single-cloud/oci/sles-sap-byos/content.yaml
smarlowucf: Ditto for the other two uses?

jgleissner: Yep.

smarlowucf:
> We'll also need `drivers` sections under `image` (below `packages` I'd say) that include `csp/oci`.

This one is confusing. If I grep for any usage of "drivers:" in the images dir there are no matches. I would assume there'd be examples to go off of? `grep -rnw images/ -e 'drivers:'`

jgleissner:
> > We'll also need `drivers` sections under `image` (below `packages` I'd say) that include `csp/oci`.
> 
> This one is confusing. If I grep for any usage of "drivers:" in the images dir there are no matches. I would assume there'd be examples to go off of? `grep -rnw images/ -e 'drivers:'`

OCI is the only image that uses `drivers`. We'll need:
```
image:
  drivers:
    - _include:
      - csp/oci
```

images/single-cloud/oci/sles-byos/15-sp3/image.yaml
```diff
@@ -0,0 +1,8 @@
+include-paths:
+  - sle15/sp3
```
jgleissner, line 2: `include-paths` should also include `sle15/sp2` and `sle15/sp1` to inherit from older SPs where applicable. Sorry, missed that earlier.

images/single-cloud/oci/sles-sap-byos/15-sp3/image.yaml
```diff
@@ -0,0 +1,8 @@
+include-paths:
+  - sle15/sp3
```
jgleissner, line 2: Same as above.

jgleissner (changes requested):

jgleissner (approved):
Looks good, thanks!

rjschwei (approved):

PR #176, opened 2022-11-21
Add 12 sp5 oci to keg

smarlowucf:
I think I updated this based on comments from the 15 SP5 PR but may have missed things. Also, not sure if there's a way we can combine some of the overlap of the 15 SP3  data?


rjschwei (approved):

jgleissner:
Looks pretty good, but I think we need `shim` in OCI, don't we?

smarlowucf:
:+1:

jgleissner (approved):

Issue #177, opened 2022-11-22
yast2-trans-en_US is defined but does not exist


mfriesenegger:
yast2-trans-en_US is defined in data/base/sle/sle15/packages.yaml but the package does not exist in SLES 15 SP2 and newer repos.

I am testing with the develop branch

jgleissner:
Fix merged.

PR #180, opened 2022-11-28
Remove unsupported packages

jgleissner:
This PR removes packages that are not supported (anymore) or were not released. Details:

- Remove rsyslog and wget from SLE Micro
- Remove yast2-trans_en_US from 15 SP2+
- Remove vlan (still supported via legacy module but simply obsolete)
- Adjust for yast2-schema successors yast2-schema-default and yast2-schema-micro
- Remove a few packages from SAPCAL that were not released
- Remove a few packages from SAPCAL SP3+ there were dropped from SLE
- Remove not-for-install system-role-sles4sap from SAP module

### DIFF
--- a/data/products/chost/sle15/packages.yaml
+++ b/data/products/chost/sle15/packages.yaml
@@ -18,12 +18,14 @@ packages:
       - nfs-client
       - open-iscsi
       - pciutils
+      - rsyslog
       - runc
       - socat
       - supportutils-plugin-suse-public-cloud
       - tcpdump
       - vim
       - vlan
+      - wget
       - wicked
   _namespace_chost_btrfsprogs:
     package:
--- a/data/base/common/sle15/packages.yaml
+++ b/data/base/common/sle15/packages.yaml
@@ -12,7 +12,6 @@ packages:
       - kdump
       - openssh
       - parted
-      - rsyslog
       - sudo
       - supportutils
       - suse-build-key
@@ -24,7 +23,6 @@ packages:
       - terminfo
       - timezone
       - udev
-      - wget
       - which
       - xfsprogs
   _namespace_common_rpm:
--- a/data/base/sle/sle15/packages.yaml
+++ b/data/base/sle/sle15/packages.yaml
@@ -50,6 +50,7 @@ packages:
       - psmisc
       - quota
       - rsync
+      - rsyslog
       - screen
       - sle-module-public-cloud-release
       - strace
@@ -61,6 +62,7 @@ packages:
       - tcsh
       - telnet
       - vim
+      - wget
       - wicked
       - yast2
       - yast2-add-on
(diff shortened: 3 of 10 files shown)


jgleissner:
Fixes https://github.com/SUSE-Enceladus/keg-recipes/issues/177.

data/products/chost/sle15/packages.yaml
```diff
@@ -18,12 +18,13 @@ packages:
       - nfs-client
       - open-iscsi
       - pciutils
+      - rsyslog
       - runc
       - socat
       - supportutils-plugin-suse-public-cloud
       - tcpdump
       - vim
-      - vlan
```
rjschwei, line 26: Please split the vlan removal into a separate PR. What replaces vlan, how does k8s setup networking and does SAP (Gardener) use anything form the vlan package are questions we need to answer first and I think we should not block the other changes while those questions get answered.

rjschwei (changes requested):
The sapcal changes also need to be split out, we need to check with SAP first

data/products/chost/sle15/packages.yaml
jgleissner: Can do, but not sure it's worth it. Quoting the vlan website (http://www.candelatech.com/~greear/vlan.html, last change 2005):

> 802.1Q VLAN code is now part of the official kernel, and has been for years and years. MAC-VLAN code has been added since around 2.6.29. It is very unlikely that you need to download anything from this site, the packages are left here for posterity's sake.
> 802.1Q VLANs and MAC-VLANs can be created with the 'ip' utility (vconfig works for vlans, but is crufty and deprecated).

jgleissner:
I've removed the changes for SAPCAL and the removal of 'vlan'. I'll create separate PRs for those.

data/products/chost/sle15/packages.yaml
rjschwei: OK, if it is part of the kernel and the kernel also delivers the user space tools, please double check, then we are OK with this drop.

jgleissner: The `vlan` package contains the `vconfig` user space tool which is about as deprecated as `ifconfig`.

rjschwei: I understand that. The question is if you are willing to cause production issues, and deal with the inevitable escalations, at SAP because vconfig is obsolete and gardner may or may not be using it? If you are willing to go down that route without checking then I am OK with merging this. From my point of view it is better to ask and then follow up with a separate PR after we know that we will not break anything.

jgleissner: Sure, that's why I dropped the removal from the PR.

jgleissner:
So, are we ok to merge?

rjschwei (approved):

PR #181, opened 2022-11-29
Add image defnitions for OCI 12 SP5

smarlowucf:


jgleissner:
The existing SLES12-SP5-OCI-BYOS is multi-build for SLES BYOS and SLES4SAP BYOS, which would be possible to implement in keg as well but is a bit more hassle. Since the OCI recipes are not intended as an update resource for the released images I think this is fine if not preferable. @rjschwei, what do you think?

rjschwei (approved):

images/single-cloud/oci/sles-byos/12-sp5/image.yaml
```diff
@@ -0,0 +1,8 @@
+include-paths:
+  - sle12/sp5
```
jgleissner, line 2: Technically we'd need to include `sle12/sp4` as well, although we could (and maybe should) alternatively merge all `sle12/sp4` specific bits into `sle12`. When keg-recipes started, we still had 12 SP3 images, so the recipes accounted for that, even though we decided to drop 12 SP3 definitions from keg-recipes, but separate configuration bits still exist which makes the include path necessary. Merging `sle12/sp4` into `sle12` would also remove a bit of clutter so unless there are objections, I'd create a PR for that.

Btw, the only thing that adding `sle12/sp4` to the OCI definition would change is the inclusion of `salt-minion`, which was added at some point to all BYOS images but it appears it was never added to OCI images, perhaps because we had already stopped refreshing them. Should we include `salt-minion` in the recipes for OCI nevertheless? I think it would make sense to keep package selection consistent across frameworks.

smarlowucf: Probably easier to add sle12/sp4 here then decide on the other changes separately?

jgleissner (approved):

rjschwei:
Merging, then we can push develop into released.

Issue #185, opened 2023-01-10
Feature: Add ConvergeCloud as Provider


killermoehre:
It was decided that SUSE will build a CHost image consumable by SAP ConvergedCloud.

This image should mirror the OpenStack one with added package `open-vm-tools`.

sandzwerg:
Image format should be streamoptimized vmdk. Which we till now build with

```
<type image="oem" filesystem="xfs" format="vmdk" firmware="uefi" kernelcmdline="plymouth.enable=0 net.ifnames=1 earlyprintk=efi,keep product_name='OpenStack Nova' consoleblank=0" bootpartition="false">
      <machine arch="x86_64" ovftype="vmware" memory="1024" HWversion="10" guestOS="sles12_64Guest"> 
<vmdisk diskmode="streamOptimized"/>
```

Probably only the `format="vmdk"`, `ovftype="vmware"` `HWversion="10"` and `<vmdisk diskmode="streamOptimized"/>` is needed

PR #186, opened 2023-01-10
Add Support for SAP Converged Cloud

killermoehre:
SAP Converged Cloud (CCloud) runs OpenStack on ESXi. With that, some kind of hybrid is necessary to make images available.

On one hand, all the OpenStack support is necessary (cloud-init mainly), on the other hand VMware-specifics like image format and other metadata needs to be considered.

Closes: #185

### DIFF
--- a/data/csp/sap-converged-cloud/sle15/packages.yaml
+++ b/data/csp/sap-converged-cloud/sle15/packages.yaml
@@ -1,11 +1,7 @@
 packages:
   _namespace_default_bootloader: Null
   _namespace_default_shim: Null
-  _namespace_openstack_bootloader:
-    package:
-      - _attributes:
-          name: grub2-arm64-efi
-          arch: aarch64
+  _namespace_openstack_bootloader: Null
   _namespace_openstack_init:
     package:
       - cloud-init
--- a/data/csp/sap-converged-cloud/sle15/packages.yaml
+++ b/data/csp/sap-converged-cloud/sle15/packages.yaml
@@ -1,0 +1,12 @@
+packages:
+  _namespace_default_bootloader: Null
+  _namespace_default_shim: Null
+  _namespace_openstack_bootloader:
+    package:
+      - _attributes:
+          name: grub2-arm64-efi
+          arch: aarch64
+  _namespace_openstack_init:
+    package:
+      - cloud-init
+      - open-vm-tools
--- a/data/csp/sap-converged-cloud/sle15/preferences.yaml
+++ b/data/csp/sap-converged-cloud/sle15/preferences.yaml
@@ -9,10 +9,10 @@ preferences:
         NON_PERSISTENT_DEVICE_NAMES: 1
         multipath: "off"
       vga: normal
-      machine:
+    machine:
+      _attributes:
+        ovftype: vmware
+        HWversion: 10
+      vmdisk:
         _attributes:
-          ovftype: vmware
-          HWversion: 10
-        vmdisk:
-          _attributes:
-            diskmode: streamOptimized
+          diskmode: streamOptimized
--- a/data/csp/sap-converged-cloud/sle15/preferences.yaml
+++ b/data/csp/sap-converged-cloud/sle15/preferences.yaml
@@ -1,0 +1,18 @@
+preferences:
+  type:
+    _attributes:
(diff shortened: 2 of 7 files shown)


data/csp/sap-converged-cloud/sle15/packages.yaml
```diff
@@ -0,0 +1,12 @@
+packages:
+  _namespace_default_bootloader: Null
+  _namespace_default_shim: Null
+  _namespace_openstack_bootloader:
+    package:
+      - _attributes:
+          name: grub2-arm64-efi
+          arch: aarch64
```
sandzwerg, line 8: As we don't have arm we can probably drop this

sandzwerg (approved):
Still needs testing as I'm not familiar with this but looks good as far as I can judge that.

sandzwerg:
@rjschwei this is the PR FYI

images/cross-cloud/sles/chost/15-sp1/content.yaml
```diff
@@ -44,9 +47,14 @@ image:
       _include:
         - csp/openstack/settings/chost
         - products/chost
+    - attributes:
+        profiles: [SAP Converged Cloud]
```
rjschwei, line 51: The profile name cannot have spaces, I suggest using `[SAP-Converged-Cloud]`. The profile definition is used to generate a `_multibuild` file for the Open Build Service. That string in turn is used when the build service executes kiw and as such would end up looking like `--profile SAP Converged Cloud`, which in turn will trip up the command line parser for kiwi due to the spaces.

rjschwei (changes requested):
Other than the profile name it looks good to me, @jgleissner could you please review this as well.

images/cross-cloud/sles/chost/15-sp1/content.yaml
jgleissner: Or maybe even just `SAPCC`.

images/cross-cloud/sles/chost/15-sp1/content.yaml
```diff
@@ -44,9 +47,14 @@ image:
       _include:
         - csp/openstack/settings/chost
         - products/chost
+    - attributes:
```
jgleissner, line 50: Should be `_attributes`.

jgleissner (changes requested):
Looks good, just a couple of minor things (see comments).

data/csp/sap-converged-cloud/sle15/preferences.yaml
```diff
@@ -0,0 +1,18 @@
...
+        NON_PERSISTENT_DEVICE_NAMES: 1
+        multipath: "off"
+      vga: normal
+      machine:
+        _attributes:
+          ovftype: vmware
+          HWversion: 10
+        vmdisk:
+          _attributes:
+            diskmode: streamOptimized
```
jgleissner, line 18: `machine` is a child of `type` not an attribute. Just move it one indentation level to the left.

images/cross-cloud/sles/chost/15-sp1/content.yaml
killermoehre: I have to go here with the SAP naming convention. So `SAP-CCloud` it is.

images/cross-cloud/sles/chost/15-sp1/content.yaml
killermoehre: Done.

data/csp/sap-converged-cloud/sle15/preferences.yaml
killermoehre: Done.

jgleissner (approved):
Looks good, thanks!

data/csp/sap-converged-cloud/sle15/packages.yaml
rjschwei: Arm is a configuration in the build service. It cannot be controlled on a flavor basis, all images built with this description will be built for x86_64 and aarch64. Therefore this is needed or we'll have build failures.

rjschwei:
@jgleissner can you please submit a PR to the Devel project with he regenerated description? Thanks

PR #187, opened 2023-01-11
Remove the no longer maintained vlan package

rjschwei:


jgleissner (changes requested):
Should go into develop first.

rjschwei:
> Should go into develop first.

Yes, my branch was created from develop, grmbl @git for not tracking this properly

jgleissner (approved):

PR #189, opened 2023-01-13
SAP-CCloud: include grub2-efi and shim packages

jgleissner:
@killermoehre @sandzwerg The SAP-CCloud profile uses firmware `uefi` but doesn't include grub2-x86_64-efi which results in build failure. This PR adds that and `shim`. Let me know if that's correct setup for SAP Converged Cloud.


rjschwei (approved):

sandzwerg:
Looks sensible Thanks

PR #190, opened 2023-01-13
Add Azure confidential VM profiles

jgleissner:
This adds Azure confidential VM profiles to SLES 15 SP4+ PAYG, SLES for SAP PAYG and SLES for SAP hardened PAYG.

I moved the SLES PAYG content definition for <= 15-SP3 into the SP2 and SP3 sub dirs and removed it from the SP4+ sub dirs (hence using the one in the parent dir), as this results in less duplication.

I didn't bother disabling the Azure CVM profile in SLES 4 SAP for < 15-SP4, as the extra profile doesn't interfere with the others. I can do that though in case we don't want to the profile defined for older SPs.


rjschwei:
@jgleissner

- still waiting for test results
- needs some adjustments to trigger kiwi to do the setup for TPM
- needs a newer version of kiwi in SLE

The changes to trigger kiwi can be pulled into the PR already, see the ModifiedImages test project in IBS

PR #196, opened 2023-02-17
Replace 'command-not-found' with 'scout-command-not-found' in 15 SP5

aosthof:
The SLES 15 SP5 Public Cloud images are failing to build due to scout [SR#289321](https://build.suse.de/request/show/289321) rename command-not-found to scout-command-not-found.

### DIFF
--- a/data/base/sle/sle15/sp5/packages.yaml
+++ b/data/base/sle/sle15/sp5/packages.yaml
@@ -1,0 +1,4 @@
+packages:
+  _namespace_sle_cnf:
+    package:
+      - scout-command-not-found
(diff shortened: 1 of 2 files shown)


jgleissner (approved):
It's really a work-around for an OBS issue IMO, but fine with me.

data/base/sle/sle15/sp5/packages.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+  _namespace_sle_cnf:
+    package:
+      - scout-command-not-found
```
rjschwei, line 4: This is actually a packaging bug. The new package should have a `Provides: command-not-found` such that this change is not necessary.

jgleissner: `scout-command-not-found` has that Provides, but it appears OBS prefers package name matches to provides matches, even if the direct match is from a lower prio repo (a bit debatable IMO).

PR #197, opened 2023-02-24
Add Azure profile for Rancher Setup

bear454:


rjschwei:
@bear454 can you please rebase. The diff has changes that have been merged and should not be there. Thanks

jgleissner:
The PR should be in the `develop` branch.

bear454:
Sorry, set the wrong merge target. Fixed, now.

jgleissner (changes requested):
Looks good in general, see other comment about the `rsyslog` issue.

bear454:
The root is that my build failed due to the inclusion of https://github.com/SUSE-Enceladus/keg-recipes/blob/develop/data/scripts/keep-default-kernel-log-level.sh#L2 ... I'm open to better solutions, but I found it problematic that the script is included even if the _rsyslog_ package isn't.

jgleissner:
Ah I see. The base `csp/azure` module configuration is intended to match standard SLES and hence assumes that `base/sle` is included, but can of course be adjusted. For example https://github.com/SUSE-Enceladus/keg-recipes/blob/develop/data/csp/azure/settings/micro/sle15/config.yaml disables some config options including the syslog option in question, `rancher-setup` could do the same.

bear454:
Thanks! I think I have the rsyslog issue resolved, now :D

jgleissner (approved):
Looks good. I'd still recommend removing `base/sle` from `config` and `archive`, even though technically no-ops.

PR #200, opened 2023-02-24
Add SLES 15 SP5 hardened PAYG image definitions

jgleissner:
Cherry-picks the addition of SLES 15 SP5 hardened PAYG image definitions from `develop`.


rjschwei:
Just cherry-picked this via he `cherry-pick` command. Looks like we need to talk about how we want to handle the transition from develop to released going forward.

PR #207, opened 2023-03-29
Remove java

rjschwei:
SAP applications no longer need our version of java. Remove it from the images.


jgleissner (approved):
FYI `java-1_8_0-ibm` is also included in `data/products/sapcal` and `data/csp/azure-baremetal`.

PR #213, opened 2023-04-25
Bug 1210579 - Include toolbox package by default in SLE Micro images

apozsuse:
Toolbox is able to work with podman (preferred) or docker ([here](https://github.com/openSUSE/microos-toolbox/blob/master/toolbox#L27)) so there is not required to change the container runtime (for toolbox at least).

I did some tests with image clouds for Micro 5.1 to 5.4 and toolbox works out of the box after installing it.

### DIFF
--- a/data/products/sle-micro/packages.yaml
+++ b/data/products/sle-micro/packages.yaml
@@ -39,6 +39,7 @@ packages:
       - systemd-presets-branding-SMO
       - systemd-default-settings-branding-SLE-Micro
       - SUSE-MicroOS-release
+      - toolbox
       - transactional-update
       - transactional-update-zypp-config
       - vim-small


jgleissner (approved):
Looks like a debug tool, so not sure whether that needs to be installed by default, but then it is tiny.

apozsuse:
Yes, this toolbox package is just a wrapper script to launch/run/enter a local debugging container (not to have to install any debug tools on the base system).

We wanted also to check this to evaluate the status in Micro images about docker/podman/containerd and k8s dropping docker support in v1.24. I'll send a mail about this. Toolbox works OK with podman or docker, so no need to change anything in this regard (for toolbox at least).

rjschwei (approved):

PR #224, opened 2023-06-14
Set SELINUX to 'enforcing' mode (bsc#1211917)

jgleissner:
Set SELINUX to enforcing mode in SLE Micro.


jgleissner:
Oops, just realized it's not supposed to be in 'enforcing' in  SLEM <5.4. I'll create a new PR.

PR #226, opened 2023-06-21
Add additional hardening to pcs-sap-hardening

brett060102:
In response to customer requests we added additional hardening rules to pcs-sap-hardening. The changes needed are:

1: Replace existing rule by rule from pcs-hardening profile    with direct use of the pcs-hardening-sap profile.

2: Add new loop to pull in additional rules from the cis_server_l1    profile. The rule list is defined by RULES_FROM_CIS variable

The new SAP hardening results in a CIS Level 1 pass rate of 81.2% compared to 66.4% for the existing SAP hardened image.


brett060102:
Image was generated from these changes in:
https://build.suse.de/package/show/home:esampson:keg-images:SLE-15-SP4/SLES15-SP4-SAP-Hardened-BYOS

QA results can be found at: https://jira.suse.com/browse/TEAM-8046

jgleissner (changes requested):
Looks ok but should go into branch `develop` first.

rjschwei:
I am not a fan of all the comments for rules we are skipping. IMHO we should track that elsewhere.

brett060102:
@jgleissner have changed PR to target develop branch.

@rjschwei have removed lines from script and put same info in the commit message. We do track the same information in confluence pages and jira's but I don't think those locations are visible outside of SUSE.

rjschwei:
@brett060102 jira and confluence are not visible outside that is correct. Our documentation for the images points to the upstream profile. If we need to make information public about what rules did not get applied and why can we put that information there?

rjschwei (approved):

brett060102:
@rjschwei Thank you. The change to the profile upstream is in process.

PR #236, opened 2023-07-26
Use oscap with pcs-hardening not remediation script

brett060102:
Use oscap with pcs-hardening not remediation script
    
    The remediation script:
    /usr/share/scap-security-guide/bash/sle15-script-pcs-hardening.sh
    
    blindly applies remediation even if it is not needed
    
    using:
    oscap xccdf eval --remediate --profile pcs-hardening /usr/share/xml/scap/ssg/content/ssg-sle15-ds.xml
    
    is better, since it will only attempt to apply remediation when needed.

### DIFF
--- a/data/scripts/pcs-hardening.sh
+++ b/data/scripts/pcs-hardening.sh
@@ -1,2 +1,5 @@
 # run image hardening script
-/usr/bin/bash -e /usr/share/scap-security-guide/bash/sle15-script-pcs-hardening.sh
+echo "run oscap -profile pcs-hardening"
+oscap xccdf eval --remediate --profile pcs-hardening /usr/share/xml/scap/ssg/content/ssg-sle15-ds.xml || {
+    echo "!!!FAILED: --profile pcs-hardening"
+}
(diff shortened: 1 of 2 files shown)


jgleissner (approved):
I can't really comment on the difference that makes for remediation, but the code looks good to me.

brett060102:
I was just looking at ensure_logrotate_activated it is failing due to the following change in 1.68 relative to 1.67

https://github.com/ComplianceAsCode/content/pull/10481

There have been a number of changes in the firewalld space. I am not sure which one is the root issue.

jgleissner:
> I was just looking at ensure_logrotate_activated it is failing due to the following change in 1.68 relative to 1.67
> 
> [ComplianceAsCode/content#10481](https://github.com/ComplianceAsCode/content/pull/10481)

I'm not familiar with how that stuff works, but one thing that catches attention is that this reads `state: "started"`. If that actually checks whether the service is started during the image build than this is likely to fail, as no services are started, obviously. Checking whether a service is enabled or not should work.

rjschwei:
Adding "do not merge for now" until the issues are sorted out. WE do not want this to accidentally end up in the generated image descriptions.

The upstream change https://github.com/ComplianceAsCode/content/pull/10481 makes the rule impossible to check during image build. That change assumes a running system.

jgleissner:
Update looks good to me.

PR #237, opened 2023-07-27
Ignore remed fail

brett060102:
Do not abort build if a SAP remediation fails
    
    Remove the /bin/fail refences from the remediation steps.
    This will allow build and image creation to procede even if
    a remedtion fails.
    
    This protects the build process from problems caused by
    security guide updates.
    
    The net effect is to make the hardened images equivalent to what
    users would see if they used the instant security guide on an
    unhardened system.


jgleissner:
PR shouldn't be in `released`. You could simply update PR https://github.com/SUSE-Enceladus/keg-recipes/pull/236 by pushing to `use-pcs-hardening`.  However, simply ignoring remediation errors doesn't seem like the right thing to do. IMO the build should fail if the remediation fails, and the remediation rules should get fixed. Otherwise introduced errors in remediation will just slip through unnoticed.

rjschwei:
I agree with Joe, the idea is that we supply hardened images and that the rules we want to apply get applied. If the application fails the image build should fail and the root cause needs to be fixed.

PR #241, opened 2023-08-16
Add explicit requirement on apache-commons-math

smarlowucf:
This prevents an unresolvable "have choice for mvn".

### DIFF
--- a/data/products/suse-manager/server/sle15/sp4/packages.yaml
+++ b/data/products/suse-manager/server/sle15/sp4/packages.yaml
@@ -4,5 +4,7 @@ packages:
       # direct include to avoid picking classpathx-mail instead of
       # javamail which does not work with spacewalk-java
       - javamail
+      # direct include to avoid unresolvable
+      - apache-commons-math
       # needed but is only recommend
       - libtcnative-1-0


smarlowucf:
Not sure if we want to handle these with prefers in build service or in the kiwi file?

smarlowucf:
Also, not sure if we can send this straight to released branch. So the images can be fixed in suse namespace. As they are breaking all SP4 publishing.

jgleissner:
We've added explicit dependencies in the past to resolve ambiguous dependencies in the build service, as this avoids change in project config. It has also bitten us a couple of times, especially if a package that we included was made obsolete by a new one. The replacement package would provide the old package of course, but with the build service apparently matching direct package names with a higher prio than provides (even from higher prio repos), this still left us including the wrong package. So I would tend to solve that via project config instead, as that doesn't add a hard dependency to the package AFAIK.

jgleissner:
Also, are you sure we want `apache-commons-math`? AFAICS `apache-commons-math3` is a SUMA specific package, so I would think that's the one required for SUMA.

smarlowucf:
> Also, are you sure we want `apache-commons-math`?

Yeah, SUMA team confirmed the preference is on `apache-commons-math`

rjschwei:
The build service change is actually just a work around. The package creating the problem should be fixed. IMHO this is a package bug. The `subscription-matcher package has` 

```Requires:       mvn(org.apache.commons:commons-math3) >= 3.2```

The SUMa team should clean this up and remove the `3` for the dependency.

For now please ask the build service team to add the prefer and please also file a SUMa bug.

PR #244, opened 2023-08-22
Add SUSE Manager Server PAYG recipes

jgleissner:
Note: This changes the source name of the SUSE Manager Server BYOS definitions and hence the `_service` files will need to be update once this is merged.


jgleissner:
I've split the Azure and EC2 profiles into llc and ltd versions. Since the PAYG and the BYOS descriptions don't have much in common anymore I moved the PAYG into an independent directory, which also has the advantage that existing `_service` files don't need to be modified.

images/cross-cloud/suse-manager/server-ondemand/content.yaml
```diff
@@ -0,0 +1,86 @@
...
+      _include:
+        - csp/ec2
+      archive:
+        - _attributes:
+            name: ec2.tar.gz
+    - _attributes:
+        type: image
+        profiles: [Azure-llc,EC2-llc]
+      package:
+        - suma-amazon-adapter-config-llc
```
aosthof: The package `suma-amazon-adapter-config-llc` is only to be included in the `EC2-llc` profile. For Azure, the equivalent package is named  `suma-azure-adapter-config-llc` and should be put into a separate `Azure-llc` profile section. Same applies for the `ltd` version of those packages. Although the Azure packages are available in Devel:PubCloud:Azure they're still empty as we need to properly setup the offer in Azure first.

From my pov this should look like:

```
    - _attributes:
        type: image
        profiles: [EC2-llc]
      package:
        - suma-amazon-adapter-config-llc
    - _attributes:
        type: image
        profiles: [EC2-ltd]
      package:
        - suma-amazon-adapter-config-ltd
    - _attributes:
        type: image
        profiles: [Azure-llc]
      package:
        - suma-azure-adapter-config-llc
    - _attributes:
        type: image
        profiles: [Azure-ltd]
      package:
        - suma-azure-adapter-config-ltd
```

data/products/suse-manager/server/ondemand/sle15/packages.yaml
aosthof, line 1: Not sure where exactly this belongs into, but we need to make sure the following `-release` packages are included:

```
        <!-- begin namespace manager_server_release -->
        <package name="SUSE-Manager-Server-release"/>
        <!-- end namespace manager_server_release -->
        <!-- begin namespace manager_server_modules -->
        <package name="sle-module-basesystem-release"/>
        <package name="sle-module-containers-release"/>
        <package name="sle-module-server-applications-release"/>
        <package name="sle-module-suse-manager-server-release"/>
        <package name="sle-module-web-scripting-release"/>
```
At least these are the corresponding `-release` package to the channels/repos a user gets when registering a BYOS version of SUMa.

data/products/suse-manager/server/ondemand/sle15/packages.yaml
```diff
@@ -0,0 +1,7 @@
+packages:
+  _namespace_manager_server_ondemand:
+    package:
+      - python-instance-billing-flavor-check
+      - python3-csp-billing-adapter
+      - python3-csp-billing-adapter-local
+      - billing-data-service
```
aosthof, line 7: IIUC we also need to include the package `csp-billing-adapter-service`

data/csp/azure/settings/suma-server/ondemand/sle15/packages.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+  _namespace_azure_suma_ondemand:
+    package:
+      - python3-csp-billing-adapter-azure
```
aosthof, line 4: Don't ask me why, but the correct package name is `python3-csp-billing-adapter-microsoft`

aosthof (changes requested):

rjschwei:
PAYG has a different motd as well

jgleissner:
Ok, updated according to comments. Thanks.

rjschwei (approved):

aosthof (approved):
LGTM now.

PR #248, opened 2023-09-14
Addtional packages in SLES 15 SP4 SAPCAL

aosthof:
As requested by Antonia Bahneva-Stoilova from SAP the following packages should be included in the SLES 15 SP4 SAPCAL images:

- xmlstarlet
- libcap-progs
- openssl
- libicu60_2
- jq

### DIFF
--- a/data/products/sapcal/sle15/sp4/packages.yaml
+++ b/data/products/sapcal/sle15/sp4/packages.yaml
@@ -1,2 +1,9 @@
 packages:
   _namespace_sapcal_libopenssl_old: Null
+  _namespace_sapcal_extra:
+    package:
+      - xmlstarlet
+      - libcap-progs
+      - openssl
+      - libicu60_2
+      - jq


data/products/sapcal/sle15/sp4/packages.yaml
```diff
@@ -1,2 +1,9 @@
 packages:
   _namespace_sapcal_libopenssl_old: Null
+  _namespace_sapcal_extra:
+    package:
+      - xmlstarlet
+      - libcap-progs
+      - openssl
```
rjschwei, line 7: openssl is already in the image. We need to verify requests.

data/products/sapcal/sle15/sp4/packages.yaml
```diff
@@ -1,2 +1,9 @@
...
+      - libicu60_2
```
rjschwei, line 8: What does `libicu60_2` provide that is not part of `libicu-suse65_1` which is shipped from the ICU package and already part of the image?

rjschwei (changes requested):

data/products/sapcal/sle15/sp4/packages.yaml
aosthof: I don't know - I explicitly asked Antonia about that and she told me "Regarding package 'libicu60_2' I had to check with my colleague who needs all those libraries and he confirmed that it is important to have exactly package 'libicu60_2' cause it`s a requirement for his product."

data/products/sapcal/sle15/sp4/packages.yaml
aosthof: I didn't find it mentioned in any of the SAPCAL related packages.yaml files, therefore I added it explicitly.

data/products/sapcal/sle15/sp4/packages.yaml
rjschwei: Saw the reply. Thanks.

aosthof:
Closing this PR to start from scratch as the package list has changed and I also want to verify that my signed commits now work.

PR #250, opened 2023-10-05
Include supportutils-plugin-salt

jgleissner:
Include package supportutils-plugin-salt in SUSE Manager Server.
Add package supportutils-plugin-susemanager explicitly to SUSE Manager Server.


aosthof (approved):
LGTM

PR #251, opened 2023-10-06
Include CSP specific billing adapter packages

jgleissner:
Include CSP specific billing adapter packages in SUSE Manager Server PAYG image definition.


jgleissner:
Also add the `base/ondemand` module so `guestregister` service gets enabled.

aosthof (approved):
LGTM

PR #252, opened 2023-10-10
Suma motd

jesusbv:
Update MOTD links for SUMa BYOS and PAYG

### DIFF
--- a/data/overlayfiles/motd-manager-server-byos/etc/motd
+++ b/data/overlayfiles/motd-manager-server-byos/etc/motd
@@ -6,8 +6,9 @@ SUSE Manager {SUMA_VERSION} Server
   \/ __ .- |
    // // @
 
-Before setting up SUSE Manager, please make sure the system's host name
-is configured to match its DNS name.
+Before setting up SUSE Manager, please make sure the system's DNS,
+network security, and storage requirements are met.
+Refer to the installation guide of the SUSE Manager documentation for details.
 
 To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
 
--- a/data/overlayfiles/motd-manager-server-byos/etc/motd
+++ b/data/overlayfiles/motd-manager-server-byos/etc/motd
@@ -13,7 +13,7 @@ To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
 
 {INCLUDES}
 Documentation: https://www.suse.com/documentation/suse_manager/
-BYOS Guide: https://documentation.suse.com/suma/{SUMA_VERSION}/en/suse-manager/specialized-guides/qs-publiccloud/overview.html
+BYOS Guide: https://documentation.suse.com/suma/{SUMA_VERSION}/en/suse-manager/specialized-guides/public-cloud-guide/byos/byos-overview.html
 Community: https://forums.rancher.com/c/suse-manager
 
 Have a lot of fun...
--- a/data/overlayfiles/motd-manager-server-byos/etc/motd
+++ b/data/overlayfiles/motd-manager-server-byos/etc/motd
@@ -1,0 +1,19 @@
+SUSE Manager {SUMA_VERSION} Server
+{OS_PRETTY_NAME} {ARCH} (64-bit)
+
+    ____
+  /@    ~-.
+  \/ __ .- |
+   // // @
+
+Before setting up SUSE Manager, please make sure the system's host name
+is configured to match its DNS name.
+
+To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
+
+{INCLUDES}
+Documentation: https://www.suse.com/documentation/suse_manager/
+BYOS Guide: https://documentation.suse.com/suma/{SUMA_VERSION}/en/suse-manager/specialized-guides/qs-publiccloud/overview.html
+Community: https://forums.rancher.com/c/suse-manager
+
+Have a lot of fun...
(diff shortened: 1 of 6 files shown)


jgleissner (changes requested):
For this to work, changes to `data/products/suse-manager/server/sle15/overlayfiles.yaml` and `data/products/suse-manager/server/ondemand/sle15/overlayfiles.yaml` are required.

data/overlayfiles/motd-manager-server-byos/etc/motd.finished
```diff
@@ -0,0 +1,30 @@
...
+ - yast scc
+
+to register the instance with SCC
+
+ - registercloudguest -r YOUR_CODE
+
+to register the instance against the SUSE update infrastructure.
+
+Without registration this instance does not have access to updates and
+security fixes.
```
jgleissner, line 23: The registration info is provided by `data/overlayfiles/motd-addon-byos`, which is used for assembling the motd file for the BYOS version. No need to include it here.

data/overlayfiles/motd-manager-server-byos/etc/motd
```diff
@@ -0,0 +1,19 @@
...
+   // // @
+
+Before setting up SUSE Manager, please make sure the system's host name
+is configured to match its DNS name.
+
+To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
+
+{INCLUDES}
+Documentation: https://www.suse.com/documentation/suse_manager/
+BYOS Guide: https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/qs-publiccloud/overview.html
```
jgleissner, line 16: That link contains a hard-coded version number. It could be replaced with https://documentation.suse.com/suma/{SUMA_VERSION}/en/suse-manager/quickstart-public-cloud/qs-publiccloud-overview.html which appears to work for 4.3 and 4.2. Links into documentation tend to break over time though.

data/overlayfiles/motd-manager-server-payg/etc/motd
```diff
@@ -12,8 +12,8 @@ is configured to match its DNS name.
 To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
 
 {INCLUDES}
-Management and Config: https://www.suse.com/suse-in-the-cloud-basics
 Documentation: https://www.suse.com/documentation/suse_manager/
-Community: https://community.suse.com/
+PAYG Guide: https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/payg/payg-overview.html
```
jgleissner, line 16: Same as in BYOS, although 4.3 is the only supported PAYG version at the moment and we don't now whether that link will work for the next SUSE Manager version, so maybe not worth changing at this point.

data/overlayfiles/motd-manager-server-byos/etc/motd
```diff
@@ -0,0 +1,19 @@
...
+   // // @
+
+Before setting up SUSE Manager, please make sure the system's host name
+is configured to match its DNS name.
+
+To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
+
+{INCLUDES}
+Documentation: https://www.suse.com/documentation/suse_manager/
+BYOS Guide: https://documentation.suse.com/suma/{SUMA_VERSION}/en/suse-manager/specialized-guides/qs-publiccloud/overview.html
```
jgleissner, line 16: That URL only seems to work for 4.3 and not for 4.2. https://documentation.suse.com/suma/{SUMA_VERSION}/en/suse-manager/quickstart-public-cloud/qs-publiccloud-overview.html works for both. But I just realized that SUSE Manager 4.2 has  only a couple of weeks of support left so perhaps this is a bit of a moot point anyway. ;)

data/overlayfiles/motd-manager-server-byos/etc/motd
```diff
@@ -0,0 +1,19 @@
...
+
+    ____
+  /@    ~-.
+  \/ __ .- |
+   // // @
+
+Before setting up SUSE Manager, please make sure the system's host name
+is configured to match its DNS name.
+
+To setup SUSE Manager please execute: sudo /sbin/yast2 susemanager_setup
```
rjschwei, line 12: I thought we had some extra comment to alert the user of the setup scrip and the disk separation?

jesusbv: Currently there is no wording for that on the MOTD

Do we want to add it or keep the wording as is ?

jgleissner: I don't think we had that in the motd. It's in the public cloud requirements section of the SUSE Manager documentation. We can add a comment with a link, of course.

jesusbv: Yes, there was not previously in the motd

Yes, I'm in favor of adding a link to that section

jesusbv: Before adding that link to the motd, one can reach that section with current 
`Documentation: https://www.suse.com/documentation/suse_manager/` -> `Installation Guide` on that page (HTML or PDF) ->  choose `Requirements` -> `Public Cloud Requirements` on that page

shall we be explicit and add it besides it can be reached through the current links ?

jesusbv: > no value in having different PAYG and BYOS motd versions wrt documentation links

Well, they point to different URLs, we can improve that in another release/version, 
main thing is BYOS and PAYG have the right URLs and get this image out 

> Neither the old nor the new version of the PAYG overview doc link works

Yes, I did test the links when they shared them as well, when I asked they said that 
that behaviour is expected and it will work soon

jgleissner: Fine with me, assuming they will work eventually. ;-)

jgleissner (approved):

rjschwei (approved):

PR #257, opened 2023-10-31
Update released

jgleissner:


rjschwei (approved):

rjschwei:
It's time to stop working around this. please fix the signing issue. This is not pleasant but since apparently we are unable to learn this simply via reminders and e-mail messages now is a good time to have a painful experience to hopefully address this once and for all.

jgleissner:
I missed that the commit wasn't signed. I don't think there's an easy way to reject unsigned commits. @aosthof Easiest way to ensure that commits are signed is to configure git to do it automatically (`git config --global commit.gpgsign true`).

jgleissner:
Also @rjschwei , as you closed the PR I'm assuming you're not happy with merging with an unsigned commit. Removing or amending the commit in `develop` would most probably require a force-push to that branch. Should we do that?

rjschwei:
I don't think a force push is required. Do one commit reverting the unsigned changes, the re-commit the SP6 changes but signed. And yes, given that we accepted unsigned commits and then I sent a message out to the ML reminding people to sign commits which was obviously not done I am not inclined to repeat that process of accepting an exception.

aosthof:
I've reverted my unsigned commit (PR#259) and resubmitted again as a signed commit (PR#260) - hope this looks better now.

PR #259, opened 2023-11-06
Revert "Add image descriptions for SLE 15 SP6"

aosthof:
This reverts my unsigned commit 43a8c1fff40b6904397a89f406fd137d97e3810b.


rjschwei (approved):

jgleissner:
Reverting the commit does not remove it, so there will still be an unsigned commit in the commit history. Granted, a reverted one, so no files from the unsigned commit remain, but we've reviewed them and they are fine. Not sure whether I see the point in doing this.

rjschwei:
I do not know if it matter for merging what is in the history. When we create the PR to released I would expect git to only look whether or not the changes in the PR are signed or not and ignore anything else. We'll see. But if you already know the answer please share.

jgleissner:
Git merges commits from one branch to another, so why wouldn't it also merge the unsigned one. Unless we cherry-pick but that comes with a ball of wax. The revert is just another commit. So the history would still be "tainted" by an unsigned commit AFAICS. But if you look up where the data comes from (e.g. with `git blame`), it would point to a signed commit if we revert and re-add, so I guess there is some value in doing that.

jgleissner (approved):

PR #267, opened 2023-12-07
Disable hardening rule aide_periodic_checking_systemd_timer

jgleissner:
Disable hardening rule aide_periodic_checking_systemd_timer as rules that require a running systemd do not work in a build environment.


jgleissner:
Do we want to replicate the remediation in the image configuration? The rule writes systemd unit files for running aide periodically and enables them.

PR #272, opened 2024-01-10
Use systemd-sysvcompat in SLE 15 SP6

jgleissner:
Use systemd-sysvcompat in SLE 15 SP6 instead of systemd-sysvinit (package rename).


aosthof (approved):
LGTM

rjschwei (approved):

PR #280, opened 2024-02-06
SLE Micro 6.0 image definitions

jgleissner:


rjschwei (changes requested):
Please drop `alp` from the path. How a distribution is put together is really meaningless for image building.

jgleissner:
We need to differentiate between `alp`, `sle12`, and `sle15` code streams, don't we? If I just remove `alp` and move the configuration bits one dir up, they will bleed into the SLE 12 and SLE 15 image descriptions.

rjschwei:
> We need to differentiate between `alp`, `sle12`, and `sle15` code streams, don't we? If I just remove `alp` and move the configuration bits one dir up, they will bleed into the SLE 12 and SLE 15 image descriptions.

But alp is a platform and will produce any number of products, sles 16, sles 16 for sap, sle-micro 6.0 and maybe others. The translation to the previous way of building the distribution would be that `sle12` and `sle15` paths would be named  `code12` and `code15`, respectively, or `common_code_base12` and `common_code_base15`. In sle we had the common code base that generated a family of products. For alp this is not true, there is no common code based. There are shared things but it is different. However, again, the way the distribution is put together should not have an effect on the image description. We continue to collect a number of package specifications etc. from various data snippets. That the underlying structure that produces the packages we want to use in the images is called ALP is immaterial. Could also be `foo` or `wehavenotquitefigureditoutyet`. 

W.r.t. bleeding into the other products, taking just one example and an easy one at that ;) we will still have `kernel-default` for example. Thus there are certainly bits that are shared across the 2 different ways how we arrange our package build in the build service, `common code base` vs. `alp`. As such I do not see this as much of an issue. I realize we want to possibly change the structure in general with reference to the `alp-testing-combined` branch. Basically I'd like us to find names that separate us from the way the names that indicate how the products are put together.

jgleissner:
Maybe I misunderstand how this will work. I assumed we'll have a common base for products derived from ALP 1.0. Maintaining a separate code base for each product seems like a lot. So my reasoning was, that image definitions have an include-path for the platform and one for the product. We could just have one for the product, but assuming that the packages and many configuration options will be the same across all ALP products, this seems like a lot of duplication. So if we have a something like an `alp/1.0` include path and a package is renamed in ALP 2.0, we only need to change it in `alp/2.0` in the respective module rather than for each affected product.

The `sle15` and `sle12` sub dirs were always supposed to be platform specifiers from my pov. Most products versions were only released for one SP so there wasn't a need to differentiate between product versions in addition to the SP versions, except for SLE Micro (which hence already has a product version specific include).

I'm not against merging SLE15 and ALP data modules and only differentiate where necessary btw. We had the strict separation between SLE12 and SLE15 to avoid changing things accidentally for the other platform, but it did cause a lot of duplication and actually duplicate work when we had to change things, since often both code streams were affected, so maybe a depart from the approach make sense. Quite a bit of SP specific stuff keg-recipes is obsolete by now or was never used (like SLE12). It would be a bit of an effort, but we could throw out everything that we don't need anymore and merge the rest together, and have sle15 specific sub dirs where necessary. This would at least get rid of the alp sub dirs for now, whether they make sense later or not.

rjschwei:
How ALP is going to be versioned is yet TBD. The only thing we know is the versioning of SLE-Micro which will be 6.0, 6.1 etc. For the SLES product Service Packages were originally supposed to go away but we do not know yet what that is supposed to look like and if the build setup moves to the component infrastructure then the platform does not get versioned at all, it is ALP and will be ALP and the components get versioned. Then each product is created via BOM that references components. 

I guess given all the uncertainty we'll keep this for now and we'll keep open to the idea that the directory tree structure is likely to change down the road as more things around ALP get nailed down.

Please pick up the file system changes from #283 before merging

rjschwei (changes requested):

data/csp/gce/alp/preferences.yaml
```diff
@@ -0,0 +1,11 @@
+preferences:
+  type:
+    _attributes:
+      format: gce
+      kernelcmdline:
+        console: ttyS0,38400n8
```
rjschwei, line 6: Please pick up the new rate setting

rjschwei (approved):

PR #281, opened 2024-02-08
Change the serial console setting for GCE

rjschwei:
A new instance type is being introduced that requires this new console setting in order to have access to the serial console. Change the value as requested.


jgleissner:
Looks fine but should go in to `develop` first.

rjschwei:
Changing merge target

PR #295, opened 2024-04-02
Add SUSE Manager Server 5.0 BYOS definition

jgleissner:
This adds the image definition and data for SUMA Server 5.0 BYOS. The image build loads the suma container file from the rpm into the podman registry and deletes the rpm afterwards.


rjschwei (approved):

aosthof (approved):
LGTM

PR #301, opened 2024-04-11
Use new podman-load-images script

jgleissner:
Use new podman-load-images script to load SUSE Manager Server container image during build.


jgleissner:
I forget to include this in the last PR.

rjschwei (approved):

PR #304, opened 2024-04-12
SUSE Manager 5.0 dependency fixes

jgleissner:
Adds two packages:

- cloud-regionsrv-client-addon-azure (is this wanted?)
- libpwquality-tools (required by cockpit-system via file dependency, which OBS can't resolve)

### DIFF
--- a/data/csp/azure/settings/suma-server/sle15/sp5/config.yaml
+++ b/data/csp/azure/settings/suma-server/sle15/sp5/config.yaml
@@ -1,8 +1,5 @@
 config:
   scripts:
     azure-default-kernel-log-level: Null
-  services:
-    azure-regionsrv-timer:
-      - regionsrv-enabler-azure.timer
   sysconfig:
     azure-set-hostname: Null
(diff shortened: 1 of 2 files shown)


data/csp/azure/settings/suma-server/sle15/sp5/packages.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+  _namespace_azure_regionsrv_client_addon:
+    package:
+      - cloud-regionsrv-client-addon-azure
```
rjschwei, line 4: We do not need this, there is no switch option from PAYG to BYOS for SUMa in Azure.

rjschwei (changes requested):
Please drop the addon

data/csp/azure/settings/suma-server/sle15/sp5/packages.yaml
jgleissner: Ok, dropped and also removed the config snippet that enables the timer.

rjschwei (approved):

PR #308, opened 2024-04-16
Fix keytable parameter

jgleissner:
Use keytable 'us' instead of 'us.map.gz'. The latter fails SLE 15 SP6 builds. Alternatively we could simply drop the parameter, since 'us' is the default anyway.


g7:
Thanks for looking at this! Please check also this conversation: https://github.com/OSInside/kiwi/pull/2515

I've tested this behaviour in 15.0, 15.3 and 15.5 and in every case systemd was okay with both the full `.map.gz` and without one (in /etc/vconsole.conf).

PR #310, opened 2024-04-18
Add cryptsetup to SLE Micro 5.5

jgleissner:
Add packages cryptsetup and device-mapper to SLE Micro 5.5, required by ignition update.


rjschwei (approved):

schaefi (approved):
yep that matches my changes :+1:

PR #311, opened 2024-04-18
Do not use boot partition

jgleissner:
Do not use boot partition for SL Micro 6.0 in Azure (bsc#1222719).


schaefi (approved):
yep makes a lot sense :+1:

PR #312, opened 2024-04-19
Add package libnsl1

jgleissner:
Add package libnsl1 to SLES for SAP 15 SP6 as required by the SAP HANA client installer (bsc#1223133).


rjschwei (approved):

rjschwei:
Sounds more like a bug in the SAP HANA client installer of not pulling it's dependencies properly. Anyway, easy enough to include.

PR #316, opened 2024-05-09
Add SUSE Manager 5.0 PAYG recipe

jgleissner:
Add SUSE Manager 5.0 PAYG recipe.
Add storage setup scripts to SUSE Manager 5.0.


rjschwei:
We want to get https://github.com/SUSE-Enceladus/susemanager-cloud-setup/pull/10 merged and then the new scripts go to the overlay directory. The old scripts do not work with SUMa in a container.

rjschwei (changes requested):
Overlay scripts are not correct

jgleissner:
Scripts updated.

rjschwei (approved):

PR #323, opened 2024-05-30
Remove python 3.11 based packages

jgleissner:
Remove python 3.11 based packages from SUSE Manager recipes.

### DIFF
--- a/data/csp/ec2/settings/suma-proxy/sle15/sp5/packages.yaml
+++ b/data/csp/ec2/settings/suma-proxy/sle15/sp5/packages.yaml
@@ -1,0 +1,8 @@
+packages:
+  _namespace_ec2_tools:
+    package:
+      - amazon-ssm-agent
+      - python3-ec2metadata
+  _namespace_ec2_image_utils:
+    package:
+      - python3-ec2imgutils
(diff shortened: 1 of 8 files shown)


data/csp/ec2/settings/suma-proxy/sle15/sp5/packages.yaml
```diff
@@ -0,0 +1,8 @@
+packages:
+  _namespace_ec2_tools:
+    package:
+      - amazon-ssm-agent
+      - python3-ec2metadata
+  _namespace_ec2_image_utils:
+    package:
+      - python3-ec2imgutils
```
rjschwei, line 8: Let's drop ec2imgutils from the SUMa images.

rjschwei (changes requested):

data/csp/ec2/settings/suma-proxy/sle15/sp5/packages.yaml
jgleissner: Ok. Should we drop the packages from 4.3 as well? It would fix the current build issues.

rjschwei (approved):

PR #327, opened 2024-06-06
Disable transactional update timer in SUMA 5.0

jgleissner:


rjschwei (approved):
We probably also want this for the proxy, please double check with the SUMa team

PR #330, opened 2024-06-12
Use kiwi-repart to grow root filesystem

jgleissner:
Use kiwi-repart to grow root filesystem in SLE Micro 5.x (bsc#1224234).

This changes the image type to `oem` and uses schema version 7.5 for the image description. We might want to consider applying that to all images as this would change the schema version to what kiwi in SLES uses natively.

### DIFF
--- a/data/products/sle-micro/5.3/preferences.yaml
+++ b/data/products/sle-micro/5.3/preferences.yaml
@@ -1,6 +1,7 @@
 preferences:
   type:
     _attributes:
+      image: oem
       kernelcmdline:
         ip: dhcp
         rd.neednet: 1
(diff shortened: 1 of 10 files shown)


data/products/sle-micro/5.3/preferences.yaml
```diff
@@ -1,6 +1,7 @@
 preferences:
   type:
     _attributes:
+      image: oem
```
schaefi, line 4: Just for the record this is an overwrite of ```image: vmx``` which is present in the defaults.yml. It makes sense here and I'm fine with it. For the "real" change to schema version 7.5 we should get rid of any ```image: vmx``` specification in the recipes though. The same applies to the schemaversion setting.

schaefi (approved):
Looks good to me :+1:

data/products/sle-micro/5.3/preferences.yaml
rjschwei: We have a lot of activity going on as such I think it is good to keep the ripples in the image descriptions to a minimum at this time and only apply the change where absolutely necessary.

rjschwei (approved):

PR #336, opened 2024-06-26
Include python-azure-agent-config-micro

jgleissner:


apozsuse (approved):
LGTM :+1:

jgleissner:
Actually, I'm not sure whether that was correct. `waagent.conf` in `python-azure-agent-micro` sets `Provisioning.Agent` 
from `cloud-init` to `disabled`, but Micro 6.0 uses cloud-init. @rjschwei, do we need the `micro` or the `server` package?

PR #339, opened 2024-07-02
Use python-azure-agent-config-server

jgleissner:
Use server instead of micro version of the python-azure-agent-config package, as the server version is intended to be used with cloud-init which SLE Micro 6 uses for instance initialization.


schaefi:
LGTM

aosthof (approved):

rjschwei:
This enables extension which is a problem. There are hundreds of extensions some of which we know, ASR for example, write to the r-o part of the filesystem on SLE-Micro. There is no way for us to document which extensions work and which extensions do not. I'll ask some questions to see where this goes.

PR #345, opened 2024-07-17
Update motd BYOS snippet

jgleissner:
Change SUSEConnect to transactional-update register in motd (bsc#1227680).
Drop motd BYOS snippet without registercloudguest info.


data/overlayfiles/motd-slfo-addon-byos/usr/lib/motd.d/20_byos_register
```diff
@@ -3,8 +3,8 @@ If the instance is not already registered, please register this instance
 using your existing SUSE entitlement.
 
 As "root" (sudo or sudo -i) use either one of the following commands:
- - SUSEConnect --url=https://scc.suse.com  -e company@example.com -r YOUR_CODE
- - yast scc
+
+ - transactional-update register -e company@example.com -r YOUR_CODE
```
rjschwei, line 7: We shold also mention th eoption to register against the update infrastructure like we do for SLE https://github.com/SUSE-Enceladus/keg-recipes/blob/released/data/overlayfiles/motd-addon-byos/etc/motd.d/10-byos-register#L10

rjschwei (changes requested):

data/overlayfiles/motd-slfo-addon-byos/usr/lib/motd.d/20_byos_register
jgleissner: The file already contains that: https://github.com/SUSE-Enceladus/keg-recipes/blob/3a4c01d473294dc479319bcf765515e73660e156/data/overlayfiles/motd-slfo-addon-byos/usr/lib/motd.d/20_byos_register

rjschwei: OK, I guess I am confused about how it gets there. Will get that information out of band. Thanks.

rjschwei (approved):

data/overlayfiles/motd-slfo-addon-byos/usr/lib/motd.d/20_byos_register
jgleissner: It was already there. There were two versions of that file, one with registercloudguest and one with out, and we used only the latter. I don't recall why. I updated the version with registercloudguest, removed the other one, and changed the Micro recipe to use the updated version.

PR #352, opened 2024-08-01
Add mount config for SLES

rjschwei:
Add libcontainers-sles-mounts package whcih provides the mount directives for the SCC credentials directories such that containers, primarily SLE BCI can access the repositories from the host.


jgleissner (approved):
LGTM.  Does the "do not merge" label still apply?

rjschwei:
Nope issues sorted out.

PR #364, opened 2024-09-17
kernel command line update

rjschwei:
Better support for rootless containers using podman. We want to make as easy as possible for our users to run containers. For rootless containers it is recommended to us cgroups v2. Set the kernel command line option for systemd to use cgroups v2.


rjschwei:
Should be to develop

PR #375, opened 2024-10-03
Add podman (bsc#1225988)

jgleissner:
This adds `podman` to SLE Micro 5.5 and also explicitly to Micro 6.0 (currently only included as a dependency of `cockpit-podman`).

Do we still keep `docker`?


rjschwei (approved):

rjschwei:
Yes we'll keep docker for now. In about 1 year all hosted k8s control planes should have dropped k8s versions that expect docker then we can drop it from our images as well.

PR #380, opened 2024-10-21
Use shim in all aarch64 recipes

jgleissner:
Drop x86_64 arch parameter from namespace default_shim.
Drop all redefinitions of shim namespace and use default one.

All maintained image descriptions should be using shim so this can be simplified.

### DIFF
--- a/data/csp/alp/defaults.yaml
+++ b/data/csp/alp/defaults.yaml
@@ -16,8 +16,7 @@ packages:
       - kernel-default
   _namespace_default_shim:
     package:
-      - _attributes:
-          name: shim
+      - shim
 preferences:
   type:
     _attributes:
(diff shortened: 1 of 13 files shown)


rjschwei:
Hmm, the signed shim for aarch64 starts with 15 SP3 and we only want to support it from 15 SP4. I tried capture that with #377. Generally I am fine with this change in keg-recipes but we have 15 SP3 that has ~1 year of life-cycle left where we might end up with potential image changes. As long as we remember not to pull this in if 15 SP3 images need regeneration, i.e. manual fiddling will be required in that case I am fine with this.

rjschwei (approved):
Note the comment

jgleissner:
Ah ok, so while this is supported for SP3 we don't want to make this change for <SP4. That's simple enough to implement in keg-recipes while still keeping most of the simplification. It's a bit on the messy side because we used to have different levels of support of shim in the earlier SLE 15 days.

rjschwei (approved):

PR #382, opened 2024-10-23
Increase default and max socket buffer sizes (bsc#1223566)

jgleissner:
This increases the default socket buffer size to 1048576 bytes as requested in bsc#1223566 and the max to 2500000 bytes (as newer kernels use as default) for all Azure images. Should this perhaps be set for all platforms to be uniform? Also, should this be limited to SLES for SAP only (the platform the bug was reported for)?

### DIFF
--- a/data/csp/azure/settings/sap/sle15/overlayfiles.yaml
+++ b/data/csp/azure/settings/sap/sle15/overlayfiles.yaml
@@ -1,0 +1,4 @@
+archive:
+  _namespace_azure_sap_sockbuf:
+    _include_overlays:
+      - sysctl-large-socket-buffer
--- a/data/csp/azure/sle15/overlayfiles.yaml
+++ b/data/csp/azure/sle15/overlayfiles.yaml
@@ -18,6 +18,3 @@ archive:
   _namespace_azure_sysconfig_network:
     _include_overlays:
       - sysconfig-network-cloud-netconfig
-  _namespace_azure_sysctl_sockbuf:
-    _include_overlays:
-      - sysctl-large-socket-buffer
--- a/data/csp/azure/sle15/overlayfiles.yaml
+++ b/data/csp/azure/sle15/overlayfiles.yaml
@@ -18,3 +18,6 @@ archive:
   _namespace_azure_sysconfig_network:
     _include_overlays:
       - sysconfig-network-cloud-netconfig
+  _namespace_azure_sysctl_sockbuf:
+    _include_overlays:
+      - sysctl-large-socket-buffer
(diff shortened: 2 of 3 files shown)


rjschwei:
Based on my understanding these settings are sensitive to the underlying network device as such generalizing them to all CSPs is probably not what we want to do. We should limit the setting to SLES4SAP as allocation of a larger buffer will trigger memory pressure elsewhere. SAP workloads are expected to run on instances with ample memory and as such this should not be a problem, however for everything else this is not the case as such we should target SLES4SAP.

rjschwei (changes requested):
Please add as SLES4SAP specific change. Thanks

jgleissner:
> Please add as SLES4SAP specific change. Thanks

Ok. After thinking about it some more, we could also only increase the max buffer size, which is 212992 bytes on SLES 15 by default (which works out to twice the size of the default weirdly because buffer sizes set via `setsockopt()` get doubled). Increasing the default buffer size would just waste kernel memory in systems that don't need it, and ironically could even lead to more packet drops if the kernel is pressed for memory.

I think it's reasonable to assume that applications that benefit from a larger receive buffer request one, and in case of UDP (which is specifically mentioned in the bug report), it very much depends on the application how to handle packet loss.

What do you think @rjschwei ?

rjschwei:
> > Please add as SLES4SAP specific change. Thanks
> 
> Ok. After thinking about it some more, we could also only increase the max buffer size, which is 212992 bytes on SLES 15 by default (which works out to twice the size of the default weirdly because buffer sizes set via `setsockopt()` get doubled). Increasing the default buffer size would just waste kernel memory in systems that don't need it, and ironically could even lead to more packet drops if the kernel is pressed for memory.
> 
> I think it's reasonable to assume that applications that benefit from a larger receive buffer request one, 


This I don't think is true or we probably wouldn't have this request.


> and in case of UDP (which is specifically mentioned in the bug report), it very much depends on the application how to handle packet loss.
> 
> What do you think @rjschwei ?

Well if we only update the max buffer size then the default would remain at whatever we build in and if the application doesn't ask for more, which I expect to be the case then we are stuck with the same problem we have today. As such I think that changing the default is the "safer" approach to meet the needs of the applications. Ultimately this is a convenience for the admin of the system and one cold look at this as a kind of pre-tuning. Which is also the reason I would narrow this to the image(s) (SLES4SAP) where we expect workloads that most likely benefit from this setting and that will run on instances where sufficient memory exists.

jgleissner:
Fair points. I feel making buffers bigger by default isn't a particularly good practice, but potentially wasting a few MB per socket on a machine designed to handle SAP workloads will probably be insignificant. I'll update the PR.

rjschwei (approved):

PR #385, opened 2024-11-04
Add SLE 15 SP7 image definitions

jgleissner:


images/cross-cloud/sle-hpc/ondemand/15-sp7/preferences.yaml
```diff
@@ -0,0 +1,22 @@
...
+      _include:
+        - csp/azure/settings/aarch64
+        - csp/azure/settings/hpc/ondemand
+    - _attributes:
+        profiles: [EC2]
+      _include:
+        - csp/ec2/settings/hpc/ondemand
+    - _attributes:
+        profiles: [GCE]
+      _include: csp/gce/settings/hpc/ondemand
```
aosthof, line 22: Not 100% sure about the syntax but does the value `csp/gce/settings/hpc/ondemand` need to be a bullet point on a new line?

images/cross-cloud/sles/byos/15-sp7/preferences.yaml
```diff
@@ -0,0 +1,22 @@
...
+      _include:
+        - csp/azure/settings/aarch64
+        - csp/azure/settings/sles/byos
+    - _attributes:
+        profiles: [EC2]
+      _include:
+        - csp/ec2/settings/sles/byos
+    - _attributes:
+        profiles: [GCE]
+      _include: csp/gce/settings/sles/byos
```
aosthof, line 22: Same as before

images/cross-cloud/sles/ondemand/15-sp7/preferences.yaml
```diff
@@ -0,0 +1,26 @@
...
+      _include:
+        - csp/azure/settings/aarch64
+        - csp/azure/settings/sles/ondemand
+    - _attributes:
+        profiles: [EC2]
+      _include:
+        - csp/ec2/settings/sles/ondemand
+    - _attributes:
+        profiles: [GCE]
+      _include: csp/gce/settings/sles/ondemand
```
aosthof, line 26: Same as before

aosthof (changes requested):
Aside from the syntax comments I made (where I'm not sure) this looks good to me.

images/cross-cloud/sle-hpc/ondemand/15-sp7/preferences.yaml
jgleissner: > Not 100% sure about the syntax but does the value `csp/gce/settings/hpc/ondemand` need to be a bullet point on a new line?

Keg supports a single string with `_include` too, but since we generally use the list format I'll change it for consistency.

aosthof (approved):
LGTM, thanks!

data/csp/sle15/sp7/defaults.yaml
```diff
@@ -0,0 +1,8 @@
+preferences:
+  type:
+    _attributes:
+      image: oem
+packages:
+  _namespace_default_oem_repart:
+    package:
+      - dracut-kiwi-oem-repart
```
rjschwei, line 8: Do we have enough time to sort out all the potential conflicts and image clean ups? We will need to disable the cloud-init disk growing mechanism for AWS and Azure and drop the stuff from the GCE images that is based on the utils package growpart-rootgrow.

So there is a lot more work than just adding the kiwi dracut definition. I am fine having the follow up work in a separate PR.

jgleissner: I didn't include the dracut Kiwi resize config part so I think the kiwi resizer should not be used (@schaefi please intervene if that is wrong). The reason why I included the package is because using image type `oem` requires it and for schema version 7.5 we need image type `oem`. And the reason for using schema version 7.5 is that it is the native version in SLE 15 and anything older than 7.4 (I think) is not supported anymore by upstream kiwi which makes generating image descriptions with recent kiwi installed a PITA.

Since the descriptions get converted into a 7.5 schema description anyway by kiwi I figured we might want to convert our image definitions in keg to use 7.5, and starting off with the SP7 ones seems like a good trial run to me.

Whether we want to use the kiwi resizer instead of the framework specific implementations is a different question.

schaefi: If you need an example kiwi description that switches the resize off for an oem build you can take a look here

* https://github.com/OSInside/kiwi/blob/main/build-tests/x86/tumbleweed/test-image-vagrant/appliance.kiwi

rjschwei: I am fine using the kiwi resize feature we just need to get rid of the other implementations then. And we'll find out if someone really cares about how we resize later. I cannot think of a way to figure this out before the switch.

schaefi: If you consider the kiwi resizer in the system you need to add a /etc/dracut.conf.d/xxx.conf which keeps the kiwi resize dracut module in the initrd such that a kernel update which calls dracut again keeps the resizer active. If we say adding this config file is something we can live with than using the kiwi resizer should be good.

It has one advantage. The kiwi resizer runs in the initrd and when the rootfs is not yet mounted. The resize therefore has no hard dependencies against other ares of the system and should not be impacted by updates on ignition, growpart, cloud-init, etc. If it proves to work once it will continue to work unless we break kiwi

rjschwei: The scheduling says that Beta2 is only 2 weeks away, Nov. 28th, which is when we usually try to have the images ready. Beta3 is Jan 23rd.

I am in favor of making the switch to the kiwi resize mechanism, ignore the self imposed beta2 deadline and give us the extra time to clean up growpart-rootgrow and the cloud-init stuff. @aosthof can you live with beta3 as the deadline to have the images ready? 

We'll need to do increased testing on the resize mechanism.

aosthof: @rjschwei yes I'm ok with that. That said we need to make sure to have images ready for testing by Beta 3 - release is planned for 2025-01-23. As the initial setup of the images in IBS takes some time until everything works as expected it would be great to have at least a first image description available in Devel:PubCloud:CrossCloud before Christmas.

jgleissner:
I've changed the SP7 recipes to use kiwi root resize instead of cloud-init/growpart-rootgrow.

data/csp/sle15/sp7/defaults.yaml
rjschwei: @schaefi could you please work with @smarlowucf to make sure we get the necessary setting to keep the repart in the initrd into our new $CSP-dracut-conf packages. Thanks.

rjschwei (approved):

smarlowucf:
@rjschwei I thought we weren't including any of the re-partition module/configs in the overlay packages?

rjschwei:
> @rjschwei I thought we weren't including any of the re-partition module/configs in the overlay packages?

Maybe I misread/didn't fully understand @schaefi 's comment about making sure the kiwi resizing code is in the initrd. 

kiwi has `dracut/modules.d/90kiwi-repart` which should live in `/usr/lib/dracut/modules.d/` and is delivered by the `dracut-kiwi-oem-repart` package. @jgleissner included the package in the image build and as such we have the resizing feature in the image. But as @schaefi points out

"""
If you consider the kiwi resizer in the system you need to add a /etc/dracut.conf.d/xxx.conf which keeps the kiwi resize dracut module in the initrd such that a kernel update which calls dracut again keeps the resizer active. If we say adding this config file is something we can live with than using the kiwi resizer should be good.
"""

Meaning we need a `$SOMENUMBER-resize.conf` in our $CSP-dracut-conf packages, but I am not certain what the content of that file should be. Otherwise the resize will only work on initial boot and will be broken once the initrd gets rebuilt by a kernel or grub update.

I think we should include that in our config packages.

But again maybe I missed something.

jgleissner:
At the moment the relevant dracut config is provided by the keg overlay module `dracut-kiwi-resize` which installs `/etc/dracut.conf.d/40-kiwi-resize.conf`. This could be moved to the dracut config package too, but at the moment we have a mix of images that use the kiwi resizer and legacy methods, so it seems easier to manage this in keg-recipes for the time being.

We could consider migrating all image descriptions to kiwi resize eventually. IMO we should probably migrate all recipes to use schema version 7.5 which is the native SLE 15 version. Since kiwi auto-migrates older schemas to that version anyway I don't see a downside of doing that, with the upside of upstream kiwi actually being able to process all image descriptions created from keg-recipes. ;) This is also required to switch to kiwi resize, which we could do at that opportunity, This would also simplify the recipes. I figured we might want to use the SP7 descriptions to trial that.

rjschwei:
@jgleissner Thanks for the details. We'll move the `40-kiwi-resize.conf` into the packages at a later time if we decide that making a whole sale switch for all maintained images is what we want to do.

PR #407, opened 2025-02-18
Do not use boot partition (bsc#1237281)

jgleissner:
Do not use boot partition for SUSE Manager 5 in Azure.


aosthof (approved):
LGTM, thanks.

rjschwei (approved):

PR #412, opened 2025-03-12
Drop instance-billing-flavor check from CHOST and Azure LI/VLI

jgleissner:
Drop package to avoid cloud-regionsrv-client being pulled into builds.


aosthof (approved):
LGTM

PR #417, opened 2025-03-28
Add SUSE Multi-Linux Manager 5.1 recipes

jgleissner:
The motd files will need an update once documentation is available.

### DIFF
--- a/data/overlayfiles/motd-mlm-proxy-5/etc/motd
+++ b/data/overlayfiles/motd-mlm-proxy-5/etc/motd
@@ -1,0 +1,26 @@
+SUSE Multi-Linux Manager {SUMA_VERSION} Proxy
+{OS_PRETTY_NAME} {ARCH} (64-bit)
+
+   ____
+  /@    ~-.
+  \/ __ .- |
+   // // @
+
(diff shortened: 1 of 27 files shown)


rjschwei (approved):

data/overlayfiles/motd-mlm-proxy-5/etc/motd
```diff
@@ -0,0 +1,26 @@
...
+{OS_PRETTY_NAME} {ARCH} (64-bit)
+
+   ____
+  /@    ~-.
+  \/ __ .- |
+   // // @
+
+To set up SUSE Multi-Linux Manager ${SUMA_VERSION} Proxy, follow the
+instructions at
+https://documentation.suse.com/suma/5.0/en/suse-manager/installation-and-upgrade/container-deployment/suma/proxy-deployment-vm-suma.html.
```
aosthof, line 11: This URL (and the others below) still refer to 5.0. But as we don't know yet if the structure will stay the same and we would have to only replace the version, I'd keep it as a reminder for now.

aosthof (approved):

PR #422, opened 2025-04-23
Use cgroup v1 for CHOST flavor builds

rjschwei:
SAP is reporting issues with the v2 cgroup setting and requesting we go back to v1. Given that we have no othet known consumers of the CHOST images we honor this request. That said the change to v1 may cause issues with k8s (https://kubernetes.io/docs/concepts/architecture/cgroups/).

### DIFF
--- a/data/products/chost/sle15/sp4/preferences.yaml
+++ b/data/products/chost/sle15/sp4/preferences.yaml
@@ -3,3 +3,4 @@ preferences:
     _attributes:
       kernelcmdline:
         cgroup.memory: Null
+        systemd.unified_cgroup_hierarchy: Null
--- a/data/csp/gdc/sle15/preferences.yaml
+++ b/data/csp/gdc/sle15/preferences.yaml
@@ -9,4 +9,3 @@ preferences:
         ci.datasource: NoCloud
         multipath: "off"
         NON_PERSISTENT_DEVICE_NAMES: 1
-        systemd.unified_cgroup_hierarchy: 1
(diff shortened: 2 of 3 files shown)


data/products/chost/sle15/sp4/preferences.yaml
```diff
@@ -3,3 +3,4 @@ preferences:
     _attributes:
       kernelcmdline:
         cgroup.memory: Null
+        systemd.unified_cgroup_hierarchy: Null
```
jgleissner, line 6: `systemd.unified_cgroup_hierarchy` isn't defined in `products/chost/`. I generally don't overwrite keys from other modules as that only works if those two modules are included in the same `_include` command and the module with the overwrite is included after the module that defined the key originally. Both conditions apply in this case so it'd be fine, just mentioning it.

jgleissner (approved):

data/products/chost/sle15/sp4/preferences.yaml
rjschwei: Well it is inherited from the higher level preference definition. Would you prefer to define a new set of preference in the chost branch? That approach would lead to 2 versions of the truth and would probably make things harder to keep in sync. This way it at least stands out as ane exception. @jgleissner ?

jgleissner: AFAICS it is only set in `csp/` (and in sub directories in `products/` other than `chost`), so that tag is not inherited from a higher level. However, picking GCE as example, the key is defined in `csp/gce/sle15/preferences.yaml` and the preferences specific image definition for GCE looks like this:
```
image:
  preferences:
    - _attributes:
        profiles: [GCE]
      _include:
        - csp/gce/settings/chost
        - products/chost
```
That means keg merges keys from the `csp/gce/settings/chost` tree including `systemd.unified_cgroup_hierarchy` and then the keys from `products/chost`, which deletes the key. I normally avoid this as it means that the order of including gains significance, and in case you include them in different namespaces it may not work as intended, but as long as that is clear I don't see an issue doing it that way (and it can even reduce some complexity).

PR #424, opened 2025-04-29
Modular motd

jgleissner:
This PR creates motd modules that use `/usr/lib/motd.d` and changes Micro and MLM 5.1 to use those (and also fixes the currently broken MLM 5.1 motds).

This makes the use of the somewhat awkward `/etc/motd` generator obsolete. Since support for `motd.d` was added to SLE 15 we could consider getting rid of all generated `/etc/motd` files. We could also consider moving the `motd.d` files to a package so they could be updated.


rjschwei (approved):

rjschwei:
We still have SLE 12 around for ~2.5 years as such I think it might make sense to keep a consistent approach and keep using the motd generator. After that, switching to `motd.d` makes sense to me. Not convinced yet that packages are the way to go. Packages would resolve the "wrong version in motd on upgrade" problem but also bring a lot of maintenance hassle with it.

jgleissner:
I agree with the maintenance hassle. The SLE 12 image descriptions we use are not generated from keg-recipes so that wouldn't be an issue.

rjschwei:
I keep forgetting about SLE 12 not being generated from keg-recipes. Let's move to `motd.d` then when time permits.

rjschwei:
Note we might need to coordinate the changes with `img-proof`

PR #425, opened 2025-04-29
Generalize billing flavor switching

rjschwei:
With version 10.4.0 of cloud regionsrv-client a general way to handle billing model switches has been implemented. This supersedes the previous Azure specific implementation. In addition a billing model change feature is being introduced in GCE. Enable billing model switched in GCE using the new implementation and switch Azure to the new implementation.


rjschwei:
Ready for review, we want to hold the merge until cloud-regionsrv-client 10.4.0 is released in SLE

jgleissner (changes requested):
Namespace `azure-regionsrv-timer` is deleted in a few product specific `settings` subdirs, namely `suma-server` (this might be wrong since Manager Server gained a PAYG option), `suma-proxy`, `sapcal`, `rancher-setup`, and `chost`. I'm guessing that the watcher is not intended to run in BYOS-only images, but even if it is, those overwrites would be superfluous with this PR.

jgleissner (approved):
LGTM

rjschwei:
cloud-regionsrv-client v 10.4.0 has been released. Merging.

PR #429, opened 2025-05-12
Use kiwi schema version 7.5

jgleissner:
Use kiwi schema version 7.5 for SL Micro 6.0.

This is needed because kiwi in SL Micro 6.0 does not support schema versions <= 7.4 anymore. The PR sets `oem-resize` to false so the kiwi dracut resizer does not get used. This would also disable it for 6.1, which I believe is still supposed to use cloud-init, but at the moment might be using the dracut resizer. We should probably also drop `dracut-kiwi-oem-repart` from 6.1 if we want to go through with this, or change this PR to enable `oem-resize` for 6.1 if not.

### DIFF
--- a/data/products/sle-micro/6.0/packages.yaml
+++ b/data/products/sle-micro/6.0/packages.yaml
@@ -19,3 +19,6 @@ packages:
       - cockpit-system
       - cockpit-tukit
       - cockpit-ws
+  _namespace_sle_micro_kiwi_resize:
+    package:
+      - dracut-kiwi-oem-repart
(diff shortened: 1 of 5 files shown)


jgleissner:
@rjschwei What do you think?

rjschwei:
@jgleissner we should switch to the kiwi resize dracut module across the board. The cloud-init resize will still run but by the time it comes around everything will be resized and thus it will just do nothing.

jgleissner:
Ok. I changed the PR to use kiwi dracut resize.

rjschwei (approved):

PR #443, opened 2025-06-19
Disable Azure Linux Agent auto update (bsc#1244933)

jgleissner:


rjschwei:
@jgleissner This works as a quick fix but we should keep this in the package like the other setting. Can you please send an SR for the package.

jgleissner:
> @jgleissner This works as a quick fix but we should keep this in the package like the other setting. Can you please send an SR for the package.

Already did: https://build.opensuse.org/request/show/1286860

PR #518, opened 2026-04-21
Update hardening recipes

jgleissner:
This PR drops a few packages from the SLES and SLES for SAP hardening recipe that were added initially but don't seem to make sense. It drops a few obsolete workarounds and enables hardening for CHOST 15 SP6+.


rjschwei:
Not sure I follow all the changes. Added a meeting to your calendar so we can discuss.

rjschwei (approved):

PR #519, opened 2026-04-23
Use new openldap2_6-client package in 16.x

aosthof:
- SLE 15 keeps openldap2-client
- SLES 16.x will use the new packag openldap2_6-client instead

### DIFF
--- a/data/base/pubcloud/_slfo/packages.yaml
+++ b/data/base/pubcloud/_slfo/packages.yaml
@@ -3,11 +3,9 @@ packages:
     package:
       - btrfsprogs
       - btrfsmaintenance
+      - openldap2_6-client
       - tpm2.0-tools
       - system-user-tss
   _namespace_base_pubcloud_slfo_podman:
     package:
       - podman
-  _namespace_base_pubcloud_slfo_extra:
-    package:
-      - openldap2_6-client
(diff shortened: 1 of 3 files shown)


data/base/pubcloud/_slfo/packages.yaml
```diff
@@ -8,3 +8,6 @@ packages:
   _namespace_base_pubcloud_slfo_podman:
     package:
       - podman
+  _namespace_base_pubcloud_slfo_extra:
```
rjschwei: What's `extra` about LDAP? LDAP is a basic authentication mechanism we want to support in our cloud images. Also the point of namespaces is that we can identify commonality across distros. So the creation of an extra namespace makes sense (spare naming issue), but then we should use that namespace consistently, i.e. use it in `_sle15` and `_slfo` and not just hack the name into the `_sle15` packages file.

rjschwei (changes requested):

data/base/pubcloud/_slfo/packages.yaml
aosthof: Well, I've added the package to the `_sle15/packages.yaml` file into the existing `_namespace_base_pubcloud_sle15` namespace. For `_slfo/packages.yaml` I was wondering whether to put it then into the existing `_namespace_base_pubcloud_slfo` or create a new one. Not knowing better I decided for the latter. But probably it should better be added to the existing `_namespace_base_pubcloud_slfo` namespace there.

aosthof: Moved the package to the generic `_namespace_base_pubcloud_slfo`

jgleissner (approved):
LGTM, thanks!

rjschwei (approved):

PR #523, opened 2026-05-04
Add script to drop non fips compliant kex algorithm

rjschwei:
CHOST images are set up to run in fips mode. The SLES ssh server configuration file includes the mlkem768x25519-sha256 key exchange algorithm, which is not fips certified. This leads to a start up failure of sshd in CHOST instances. Remove the offending kex algorithm from the config.

### DIFF
--- a/data/products/_sles/chost/16.0/config.yaml
+++ b/data/products/_sles/chost/16.0/config.yaml
@@ -1,3 +1,5 @@
 config:
   scripts:
     products_sles_selinux: Null
+    fips_kex:
+      - ssh-drop-non-fips-kex-algo
--- a/data/scripts/ssh-drop-non-fips-kex-algo.sh
+++ b/data/scripts/ssh-drop-non-fips-kex-algo.sh
@@ -1,0 +1,4 @@
+# Drop Key Exchange Algorithm that is not fips approved for ssh
+# bsc#1259495, bsc#1262555
+sed -i -e 's/mlkem768x25519-sha256,*//g; s/,,/,/g; s/,$//g'
+    /etc/crypto-policies/back-ends/opensshserver.config


rjschwei:
@jgleissner I could not get keg to cooperate with me as such I was not able to generate a description and test this change. Was using 

```
keg --version
2.2.1
```

No matter what I did I got the help message back.

jgleissner:
How did you run keg?

data/scripts/ssh-drop-non-fips-kex-algo.sh
```diff
@@ -0,0 +1,4 @@
+# Drop Key Exchange Algorithm that is not fips approved for ssh
+# bsc#1259495, bsc#1262555
+sed -i -e 's/mlkem768x25519-sha256,*//g; s/,,/,/g; s/^,//g; s/,$//g'
```
jgleissner: The `s/^,//g` seems a bit superfluous because no line should start with `,`, there should always be a keyword. The first expression already deletes the comma so even if `mlkem768x25519-sha256` is the first algorithm of the list it wouldn't leave a leading comma. But it won't to any harm.

jgleissner (approved):

data/products/_sles/chost/16.0/config.yaml
```diff
@@ -1,3 +1,5 @@
 config:
   scripts:
     products_sles_selinux: Null
+    fips_kex:
+      - ssh-drop-non-fips-kex-algo.sh
```
jgleissner: The `.sh` needs to be dropped. Keg appends that automatically. Sorry I missed that earlier.

jgleissner (changes requested):

data/products/_sles/chost/16.0/config.yaml
rjschwei: I knew that and then obviously put the .sh extension there anyway. grumble.

data/scripts/ssh-drop-non-fips-kex-algo.sh
rjschwei: True that, fixed.

jgleissner (approved):

PR #530, opened 2026-05-21
Include parted in all images

jgleissner:
Add `parted` to common package selection so it is included in all images (currently missing from 16.x).


smarlowucf (approved):
LGTM

PR #531, opened 2026-05-21
Include which in all images

smarlowucf:
Add which to common package selection so it is included in all images (currently missing from 16.x).

### DIFF
--- a/data/base/common/_sle15/packages.yaml
+++ b/data/base/common/_sle15/packages.yaml
@@ -26,6 +26,7 @@ packages:
       - terminfo
       - timezone
       - udev
+      - which
       - xfsprogs
   _namespace_base_common_netsys:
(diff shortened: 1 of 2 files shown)


rjschwei (approved):

data/base/common/_sle15/packages.yaml
```diff
@@ -26,7 +26,6 @@ packages:
       - terminfo
       - timezone
       - udev
-      - which
```
jgleissner, line 29: This will remove `which` from SLE15 based recipes as the common package selection is separate for SLE15 (`_namespace_base_common` is overwritten here).

jgleissner (changes requested):

jgleissner (approved):
Thx!

PR #533, opened 2026-06-02
Set up FIPS mode in SLES CHOST 16.x

jgleissner:
Enable FIPS crypto policy.
Drop unsupported AppArmor packages.

### DIFF
--- a/data/base/pubcloud/_sles/chost/config.yaml
+++ b/data/base/pubcloud/_sles/chost/config.yaml
@@ -11,7 +11,6 @@ config:
       - chost-strip-image
     base_pubcloud_chost_scripts_fips:
       - enable-fips-mode
-      - ssh-drop-non-fips-kex-algo
   services:
     base_pubcloud_chost_services:
       - containerd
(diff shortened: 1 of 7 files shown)


data/base/pubcloud/_sles/chost/config.yaml
```diff
@@ -10,7 +10,8 @@ config:
     base_pubcloud_chost_scripts:
       - chost-strip-image
     base_pubcloud_chost_scripts_fips:
-      - apparmor-enable-complain
+      - enable-fips-mode
+      - ssh-drop-non-fips-kex-algo
```
rjschwei: This should theoretically no longer be necessary as the fips policies should not contain any non-fips approves algorithms. The `ssh-drop-non-fips-kex-algo` removes a non fips algorithm form the config such that sshd does not abort.

jgleissner: Great. Always nice to get rid of a work-around.

rjschwei (approved):

PR #542, opened 2026-06-18
Add libnsl to sles 16.x sapcal images

aosthof:
This has been requested by Kalin Mikov from the SAP Cloud Application Library team and is tracked in https://jira.suse.com/browse/PCT-1827


jgleissner (changes requested):
Needs to be opened against `main`.

PR #547, opened 2026-07-03
Drop duplicate GCE package section

jgleissner:
Drop GCE from the packages section of extra profiles. This is already covered by 'gce-base' and only results in duplicates.


aosthof (approved):
LGTM

gubasso (approved):
LGTM

PR #550, opened 2026-07-27
Enable tiertune in CHOST

jgleissner:


data/platforms/csp/ec2/_sles/chost/15-sp4/packages.yaml
```diff
@@ -1,2 +1,3 @@
 packages:
   _namespace_platforms_ec2_sle15: Null
+  _namespace_platforms_ec2_tiertune:
```
rjschwei: `Null` missing?

data/platforms/csp/gce/_sles/chost/15-sp6/preferences.yaml
```diff
@@ -0,0 +1,9 @@
+preferences:
+  type:
+    _attributes:
+      kernelcmdline:
+        add_efi_memmap udev.children-max: 512
```
rjschwei: Do these have to combine on one line since `add_efi_memmap` has no value? It looks a bit odd.

jgleissner: Syntax for valueless params is using empty list as value (as YAML keys must have something). Thanks for spotting, fixed. I was rushing the PR. Sorry for the avoidable errors.

data/platforms/csp/gce/_sles/chost/preferences.yaml
```diff
@@ -0,0 +1,9 @@
+preferences:
+  type:
+    _attributes:
+      kernelcmdline:
+        add_efi_memmap udev.children-max: 512
```
rjschwei: These were separated for the 15-sp6 config should they get separated here as well?

jgleissner: I don't even have an excuse for that one. :|

rjschwei (approved):

PR #556, opened 2026-08-06
Change googel guest agent service enablement

rjschwei:
Google is changing the transition plan to the new code base. Initially we will need both services google-guest-agent and google-guest-agent-manger running in parallel. While the google-guest-agent-manger does not yet exist adding it now ensures it does not forgotten later when we add the new package to he image descriptions.


jgleissner (approved):
LGTM. Kiwi's baseInsertService function, which is used in `config.sh` for enabling services, hides errors, so merging this won't break the builds.

PR #562, opened 2026-08-20
Set /etc/machine-id to uninitialized

jgleissner:
Set /etc/machine-id to uninitialized so first boot of instance is actually considered first boot by systemd.


rjschwei (approved):

schaefi (approved):
LGTM 👍

PR #563, opened 2026-08-21
Fix motd for SLES for SAP Applications

jgleissner:
Append "for SAP Applications" to product name in OS header in motd.


apozsuse (approved):
lgtm. Thx!

PR #564, opened 2026-08-26
Fix some issues with motd

jgleissner:
- Provide product ID and version as variables for motd snippets
- Replace product specific documentation links motd snippets with generic one
- Include documentation links in all products
- Include software management hint in SLE HPC PAYG motd
- Include registration motd hint in SLE HPC BYOS motd
- Install SLE_HPC-release instead of sles-release in SLE HPC recipes


rjschwei (approved):
Not sure that the change around product version works across all variants of /etc/os-release, but trust that you checked the current maintained versions 15 SP4 through 16.1

